PR #205, opened 2025-10-21
Do not cache device names

farshidtz:
Closes canonical/inference-snaps#151


jpm-canonical:
> embed the pci-ids in every snap rather than relying on the remote server.

That's what we had before. Reference:
https://github.com/canonical/inference-snaps-cli/commit/a7a2ee05ee085fca4db99d7f5ea05ae56046749e

PR #214, opened 2025-10-24
Timeout external tool calling

jpm-canonical:


pkg/hardware_info/pci/intel/gpu.go
```diff
@@ -29,12 +34,42 @@ func vRam(device types.PciDevice) (*uint64, error) {
...
+		if cmdOutput.err != nil {
+			if len(cmdOutput.output) == 0 {
+				return nil, cmdOutput.err
+			} else {
+				return nil, fmt.Errorf("%s: %s", cmdOutput.err, bytes.TrimSpace(cmdOutput.output))
+			}
+		}
+		data = cmdOutput.output
+	}
+
```
farshidtz: This could be moved to a function and reused.

jpm-canonical:
The below code proves that using the `WaitDelay` in the standard library solves the child process killing problem too:

```
	ctx := context.Background()
	cmdContext, cancel := context.WithTimeout(ctx, 10*time.Second)
	defer cancel()

	command := exec.CommandContext(cmdContext, "sh", "-c", "sleep 30")
	command.WaitDelay = 1 * time.Second

	data, err := command.Output()
	if err != nil {
		return nil, err
	}
```

```
=== RUN   TestGetVram
    gpu_test.go:41: signal: killed
--- FAIL: TestGetVram (11.01s)
```

farshidtz:
Superseded by #216

PR #316, opened 2026-03-31
Create symlinks for component layouts

mrmara:
Add possibility to create symbolic links from `component.yaml` usign layouts syntax:
```
layout:
  target:
    symlink: source
```


cmd/cli/common/engine.go
```diff
@@ -69,19 +74,21 @@ func EngineComponentSettings(ctx *Context) ([]ComponentSettings, error) {
...
 	componentsDir, found := os.LookupEnv("SNAP_COMPONENTS")
 	if !found {
 		return fmt.Errorf("SNAP_COMPONENTS env var not set")
 	}
 
 	for _, settings := range settingsCollection {
+
+		// Set component path env var for expansion
+		componentPath := filepath.Join(componentsDir, settings.componentName)
+		if err := os.Setenv(componentEnv, componentPath); err != nil {
```
farshidtz, line 86: With this change, the COMPONENT variable is no longer unset after its use. This variable is only needed for the env expansion. It should not leak into the process's environment because it will not make sense in that context with multiple components per engine. It is valid from the context of each component.

cmd/cli/common/engine.go
```diff
@@ -91,31 +98,90 @@ func LoadEngineEnvironment(ctx *Context) error {
...
+				}
+				err := os.Symlink(target, link)
+				if err != nil {
+					return fmt.Errorf("error creating symlink from %q to %q: %v", link, target, err)
+				}
+			}
+		}
 	}
 
 	return nil
```
farshidtz, line 144: Return the unload function. This is a common pattern in Go, where functions return a pointer to another function responsible for the cleanup.

```suggestion
	return unloadEngineEnvironmentFromSettingsCollection(settingsCollection)
```

The error returned by unloadEngineEnvironmentFromSettingsCollection should be handled somewhere. If here, consider:
```go
return func(){
  if err := unloadEngineEnvironmentFromSettingsCollection(settingsCollection); err != nil {
    fmt.Println(...)
  }
}
```

cmd/cli/common/engine.go
```diff
@@ -91,31 +98,90 @@ func LoadEngineEnvironment(ctx *Context) error {
...
+				return fmt.Errorf("invalid env var %q", kv)
+			}
+			k := parts[0]
+
+			err := os.Unsetenv(k)
+			if err != nil {
+				return fmt.Errorf("error unsetting %q: %v", k, err)
+			}
+		}
+	}
```
farshidtz: This shouldn't be needed because unlike symlinks, environment variables do not persist.

cmd/cli/common/engine.go
```diff
@@ -91,31 +98,90 @@ func LoadEngineEnvironment(ctx *Context) error {
...
+	// remove the symlinks created for the engine components
+	for _, settings := range settingsCollection {
+		for layoutPath := range settings.Layout {
+			layoutPath = os.ExpandEnv(layoutPath)
+			if err := os.RemoveAll(layoutPath); err != nil {
+				return fmt.Errorf("error removing symlink %q: %v", layoutPath, err)
```
farshidtz: The caller isn't capturing the returned error. This should either log the error and return nothing or the caller should capture it and do so.

cmd/cli/commands/run.go
```diff
@@ -54,13 +54,15 @@ func (cmd *runCommand) run(_ *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("error loading engine environment: %v", err)
 	}
-
+	defer common.UnloadEngineEnvironment(cmd.Context)
```
farshidtz: Add more cohesion:
```go
clean, err := common.LoadEngineEnvironment(cmd.Context)
if err != nil {
  return fmt.Errorf("error loading engine environment: %v", err)
}
defer clean()
```

cmd/cli/commands/run.go
```diff
@@ -54,13 +54,15 @@ func (cmd *runCommand) run(_ *cobra.Command, args []string) error {
...
 	path := args[0]
 
 	execCmd := exec.Command(path)
 	execCmd.Stdout = os.Stdout
 	execCmd.Stderr = os.Stderr
-	return execCmd.Run()
+	err = execCmd.Run()
+
+	return err
```
farshidtz: This change is redundant.

cmd/cli/common/engine_test.go
```diff
@@ -0,0 +1,181 @@
...
+				"/NOT/EXISTENT": {
+					Symlink: "$SNAP_COMPONENTS/dummy-component-2/non_existent_file.txt",
+				},
+			},
+			Environment: []string{
+				"TEST_ENV_VAR=test",
+			},
+		},
+	}
+}
```
farshidtz: These look over-engineered. I suggest moving them to the tests that need them.

For `invalidVarTestSettingsCollection(symlinkPath string)`, the input is never set.

farshidtz (changes requested):
Thanks. Please see my comments inline.

cmd/cli/commands/run.go
```diff
@@ -49,12 +51,29 @@ func (cmd *runCommand) run(_ *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("error loading engine environment: %v", err)
 	}
+
 	defer clean()
```
farshidtz, line 55: I thought we've established that deferred statements aren't called because `execCmd.Run()` exits on stop signals instead of returning.

Can this work? (remove lines 63-76)
```
    sigChan := make(chan os.Signal, 1)
    signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
    go func() {
      <-sigChan
      clean()
    }
```

farshidtz: I retract this suggestion, as it can lead to cleanup before the child process tree exits. This might be okay now because these are only symlinks, but there is nothing preventing the clean function from evolving independently.

cmd/cli/common/engine.go
```diff
@@ -19,10 +19,16 @@ const (
...
+
 type ComponentSettings struct {
-	componentName string
-	Servers       map[string]map[string]string `yaml:"servers"`
-	Environment   []string                     `yaml:"environment"`
+	componentName  string
+	Servers        map[string]map[string]string `yaml:"servers"`
+	Environment    []string                     `yaml:"environment"`
+	Layout         map[string]ComponentLayout   `yaml:"layout"`
+	ExpandedLayout map[string]ComponentLayout
```
farshidtz: Keep this internal:
```suggestion
	expandedLayout map[string]ComponentLayout
```

cmd/cli/common/engine.go
```diff
@@ -69,19 +75,21 @@ func EngineComponentSettings(ctx *Context) ([]ComponentSettings, error) {
...
 		return fmt.Errorf("SNAP_COMPONENTS env var not set")
 	}
 
-	for _, settings := range settingsCollection {
+	for i, settings := range settingsCollection {
+
+		// Set component path env var for expansion
+		componentPath := filepath.Join(componentsDir, settings.componentName)
+		if err := os.Setenv(componentEnv, componentPath); err != nil {
+			return fmt.Errorf("error setting %q: %v", componentEnv, err)
```
farshidtz: Consistent with #319:
```suggestion
			return fmt.Errorf("setting env %q: %v", componentEnv, err)
```

Please drop `error ` from other error messages as well.

cmd/cli/common/engine.go
```diff
@@ -91,31 +99,83 @@ func LoadEngineEnvironment(ctx *Context) error {
...
+
+				// Ensure the parent directory for the link exists
+				if err := os.MkdirAll(filepath.Dir(link), 0755); err != nil {
+					return fmt.Errorf("error creating directory for symlink %q: %v", link, err)
+				}
+
+				// Remove existing file, symlink, or directory if it exists
+				if err := os.RemoveAll(link); err != nil {
+					return fmt.Errorf("error removing existing path %q: %v", link, err)
+				}
```
farshidtz, line 133: Removing existing files and directories can be catastrophic since the snap can get access to the home and other directories on the host (depending on the packaging). A developer can mistakenly add a link on a top level system directory and lose everything, in dev mode or via the snap interfaces.

I think we should put a limit on this to only the [data locations](https://snapcraft.io/docs/reference/administration/data-locations/) within the snap. Given that the links are recreated every time this runs, I don't see any value in supporting all writable locations. How about rejecting anything other than `/tmp` as link path suffix when creating the links? `/tmp` inside a snap translates to `/tmp/snap-private-tmp/snap.<snap-instance>/tmp` on the host.

cmd/cli/common/engine.go
```diff
@@ -91,31 +99,83 @@ func LoadEngineEnvironment(ctx *Context) error {
...
+				err := os.Symlink(target, link)
+				if err != nil {
+					return fmt.Errorf("error creating symlink from %q to %q: %v", link, target, err)
+				}
```
farshidtz: Consistent with #319:
```suggestion
					return fmt.Errorf("creating symlink from %q to %q: %v", link, target, err)
				}
```

cmd/cli/common/engine.go
```diff
@@ -91,31 +99,83 @@ func LoadEngineEnvironment(ctx *Context) error {
...
+func unloadEngineEnvironmentFromSettingsCollection(settingsCollection []ComponentSettings) error {
+
+	// remove the symlinks created for the engine components
+	for _, settings := range settingsCollection {
+		for layoutPath, _ := range settings.ExpandedLayout {
+			if err := os.RemoveAll(layoutPath); err != nil {
+				fmt.Printf("Error removing symlink %q: %v\n", layoutPath, err)
+				return err
+			}
+		}
```
farshidtz: Return error and drop extra blank identifier (_):
```suggestion
		for layoutPath := range settings.ExpandedLayout {
			if err := os.RemoveAll(layoutPath); err != nil {
				return fmt.Errorf("removing symlink %q: %v", layoutPath, err)
			}
		}
```

cmd/cli/common/engine.go
```diff
@@ -91,31 +99,83 @@ func LoadEngineEnvironment(ctx *Context) error {
...
+
+	err = loadEngineEnvironmentFromSettingsCollection(settingsCollection)
+
+	return func() error {
+		if err := unloadEngineEnvironmentFromSettingsCollection(settingsCollection); err != nil {
+			fmt.Printf("error unloading engine environment: %v\n", err)
+		}
+		return nil
+	}, err
+}
```
farshidtz: Returned error not used. 
Make the printed error a warning to stderr, based on #319. We only print "Error: xxx" for fatal errors. 
```suggestion
func LoadEngineEnvironment(ctx *Context) (func(), error) {
	settingsCollection, err := EngineComponentSettings(ctx)
	if err != nil {
		return nil, fmt.Errorf("error loading engine component settings: %v", err)
	}

	err = loadEngineEnvironmentFromSettingsCollection(settingsCollection)

	return func() {
		if err := unloadEngineEnvironmentFromSettingsCollection(settingsCollection); err != nil {
			fmt.Fprintf(os.Stderr, "Warning: failed to unload engine environment: %v\n", err)
		}
		return
	}, err
}
```

farshidtz (approved):
LGTM. Thanks for optimizing the env processing as well.

PR #319, opened 2026-04-09
Remove duplication from error chains

jpm-canonical:
* Remove duplication from error chains
* Warnings are printed to stderr
* Abbreviations are all lower case
* Each level in the chain adds a prefix


errors-summary.md
```diff
@@ -0,0 +1,499 @@
+# Errors Summary
```
jpm-canonical: Remove this file before merging

cmd/cli/common/errors.go
```diff
@@ -4,4 +4,6 @@ import "errors"
 
 var (
 	ErrPermissionDenied = errors.New("permission denied, try again with sudo")
+	ErrGetActiveEngine  = errors.New("looking up active engine")
```
farshidtz: This is named after an operation and doesn't sound like a type of error. I would remove this error type and use the string where applicable. 

Extended explanation:
I see that the error is used as a prefix and returned as wrapped to the caller. This "invites" the caller to expect and handle this error type. The error type is never checked and handled in a distinct way right now, and in my opinion, a type is misused here to avoid replicating a string. This is acceptable when it is only consumed for its string value. The issue is that "indirect callers of GetActiveEngine()" may rely on it, but the direct caller may not use or wrap it. In other words, the error belongs to GetActiveEngine() but is thrown by its caller.

pkg/hardware_info/pci/intel/gpu.go
```diff
@@ -49,12 +49,12 @@ func vRam(device types.PciDevice) (*uint64, error) {
 
 	data, err := command.Output()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("querying clinfo: %v", err)
```
farshidtz: Use command line terminology: 
```suggestion
		return nil, fmt.Errorf("executing clinfo: %v", err)
```

pkg/hardware_info/pci/intel/gpu.go
```diff
@@ -49,12 +49,12 @@ func vRam(device types.PciDevice) (*uint64, error) {
...
 	}
 
 	clinfo, err := parseClinfoJson(data)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse clinfo json: %w", err)
+		return nil, fmt.Errorf("parsing clinfo response: %w", err)
```
farshidtz:
```suggestion
		return nil, fmt.Errorf("parsing clinfo output: %w", err)
```

pkg/hardware_info/pci/nvidia/gpu.go
```diff
@@ -49,7 +49,7 @@ func vRam(device types.PciDevice) (*uint64, error) {
 	*/
 	output, err := nvidiaSmi("--id="+device.Slot, "--query-gpu=memory.total", "--format=csv,noheader")
 	if err != nil {
-		return nil, fmt.Errorf("error executing nvidia-smi: %s", err)
+		return nil, fmt.Errorf("querying nvidia-smi: %s", err)
```
farshidtz:
```suggestion
		return nil, fmt.Errorf("executing nvidia-smi: %s", err)
```

pkg/hardware_info/pci/nvidia/gpu.go
```diff
@@ -77,7 +77,7 @@ func computeCapability(device types.PciDevice) (*string, error) {
 	// nvidia-smi --query-gpu=compute_cap --format=csv,noheader
 	output, err := nvidiaSmi("--id="+device.Slot, "--query-gpu=compute_cap", "--format=csv,noheader")
 	if err != nil {
-		return nil, fmt.Errorf("error executing nvidia-smi: %s", err)
+		return nil, fmt.Errorf("querying nvidia-smi: %s", err)
```
farshidtz:
```suggestion
		return nil, fmt.Errorf("executing nvidia-smi: %s", err)
```

pkg/hardware_info/pci/pci_devices.go
```diff
@@ -25,11 +25,11 @@ func Devices(friendlyNames bool) ([]types.PciDevice, error) {
 
 	hostLsPciData, err := hostLsPci()
 	if err != nil {
-		return nil, fmt.Errorf("error getting host lspci data: %v", err)
+		return nil, fmt.Errorf("querying host lspci: %v", err)
```
farshidtz:
```suggestion
		return nil, fmt.Errorf("executing lspci: %v", err)
```

pkg/hardware_info/pci/pci_devices.go
```diff
@@ -25,11 +25,11 @@ func Devices(friendlyNames bool) ([]types.PciDevice, error) {
...
 	}
 	devices, err := ParseLsPci(hostLsPciData, friendlyNames)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing lspci data: %v", err)
+		return nil, fmt.Errorf("parsing lspci response: %v", err)
```
farshidtz:
```suggestion
		return nil, fmt.Errorf("parsing lspci output: %v", err)
```

pkg/hardware_info/pci/pci_devices.go
```diff
@@ -42,7 +42,7 @@ func Devices(friendlyNames bool) ([]types.PciDevice, error) {
 func DevicesFromRawData(lspciData string, friendlyNames bool) ([]types.PciDevice, error) {
 	devices, err := ParseLsPci(lspciData, friendlyNames)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing lspci data: %v", err)
+		return nil, fmt.Errorf("parsing lspci response: %v", err)
```
farshidtz:
```suggestion
		return nil, fmt.Errorf("parsing lspci output: %v", err)
```

farshidtz (changes requested):
Thanks, I really like the overall simplification and improved traceability with error chaining adjustments. 

Most of my comments are redundant and about command execution terminology (exec & output vs query and response). I think "query" and "response" are perceived more as network or DB terms, while "exec/run" and "output" make it clear that a local program is being started, setting up more meaningful context before the rest of the error message.

cmd/cli/commands/prune-cache.go
```diff
@@ -47,17 +47,17 @@ func (cmd *pruneCacheCommand) run(_ *cobra.Command, _ []string) error {
...
+		return fmt.Errorf("%w: %w", common.ErrGetActiveEngine, err)
 	}
 	activeEngineManifest, err := engines.LoadManifest(cmd.EnginesDir, activeEngine)
 	if err != nil {
 		if errors.Is(err, engines.ErrManifestNotFound) {
 			if cmd.Verbose {
 				fmt.Println(err)
 			}
-			return fmt.Errorf("No active engine found")
+			return common.ErrNoActiveEngine
```
farshidtz: ErrNoActiveEngine (no active engine set) is used here in place of "active set but not found". These are two distinct errors.

cmd/cli/common/errors.go
jpm-canonical: I get what you are saying. This string is added to the error in multiple places, so I guess a string constant would be a better approach.

cmd/cli/commands/prune-cache.go
jpm-canonical: I've changed it to: ```active engine manifest not found```

I'm not even sure we should be returning an error here. If prune is called, with no engine being active, the prune should not stop, but remove all components. Issue opened for it: https://github.com/canonical/inference-snaps/issues/264

cmd/cli/commands/prune-cache.go
```diff
@@ -191,7 +191,7 @@ func (cmd *pruneCacheCommand) printComponentsAndConfirm(componentsWithEngines ma
 
 		componentSizes, err := snap_store.ComponentSizes()
 		if err != nil {
-			fmt.Printf("Warning: unable to get component sizes: %v\n", err)
+			fmt.Printf("Warning: unable to query component sizes: %v\n", err)
```
farshidtz:
```suggestion
			fmt.Fprintf(os.Stderr, "Warning: unable to query component sizes: %v\n", err)
```

farshidtz (approved):

errors-summary.md
farshidtz: Please don't forget to remove this.

PR #320, opened 2026-04-10
Use utils function to detect terminal

jpm-canonical:
We have a function in utils to detect if the current session is a terminal. It is however only used for `get` and the code is replicated in `prune-cache` and `use-engine`.

### DIFF
--- a/cmd/cli/commands/use-engine.go
+++ b/cmd/cli/commands/use-engine.go
@@ -15,7 +15,6 @@ import (
 	"github.com/canonical/inference-snaps-cli/pkg/snap_store"
 	"github.com/canonical/inference-snaps-cli/pkg/utils"
 	"github.com/spf13/cobra"
-	"golang.org/x/term"
 )
 
 type useEngineCommand struct {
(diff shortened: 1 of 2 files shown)


cmd/cli/commands/use-engine.go
```diff
@@ -329,7 +328,7 @@ func (cmd *useEngineCommand) installMissingComponents(engine *engines.Manifest)
 	}
 
 	// Only ask for confirmation if it is an interactive terminal
-	if !cmd.assumeYes && term.IsTerminal(int(os.Stdin.Fd())) {
+	if !cmd.assumeYes && utils.IsTerminalOutput() {
```
farshidtz, line 331: I recall that this checked Stdin instead of Stdout for a reason. I can't remember the reason.

https://github.com/canonical/inference-snaps-cli/blob/123cd9c33b6cfd76ca73fa6d16650d28afed94fc/pkg/utils/utils.go#L111-L113

farshidtz: I did some further digging. The function in utils [was implemented](https://github.com/canonical/inference-snaps-cli/commit/7cd93850725ca14c410fbe0dedf4fe9f4e365151) for the `show-engine` command that can be piped to get the engine name. The function was used to detect if there is an stdout to decide if highlighting should be applied. This function is no longer used there.

farshidtz: I am not able to think of a situation where stdin is a terminal but not stdout. So we should be able to use `term.IsTerminal(int(os.Stdout.Fd()))` and `term.IsTerminal(int(os.Stdin.Fd()))` interchangeably in this CLI.

farshidtz (approved):
A good improvement. However, I am in favor of removing the terminal check in both cases and relying on --assume-yes flag (missing for prune-cache).

jpm-canonical:
> removing the terminal check in both cases and relying on --assume-yes flag (missing for prune-cache).

Adding an issue for it:
https://github.com/canonical/inference-snaps/issues/265

PR #321, opened 2026-04-10
Only show component size warning in verbose mode

jpm-canonical:
Not being able to look up the component size is not critical, and a user should not even be notified if it failed. 

The warning can be very verbose, containing http error codes. Here we change the warning to only be shown in verbose mode.


farshidtz (approved):
Thanks. This should resolve https://github.com/canonical/inference-snaps/issues/241

PR #322, opened 2026-04-13
Add strict checks in layout creation to avoid accidental file removal

farshidtz:
Here is a fun story: I was trying to improve the following test which doesn't check nil error (directory created successfully outside temp). At some point I wanted to make `os.MkdirAll` fail by setting a smart path on line 142. It had to start with `/tmp` to bypass the prefix check but needed to fail when creating the directory, so I came up with: `/tmp/../`! In a matter of seconds, all my user-owned files were gone. I made this mistake as I didn't expect a test named "TestCannotCreateDir" end up making calls to `os.RemoveAll`. The good thing is that I really needed the disk space. 🙂

https://github.com/canonical/inference-snaps-cli/blob/212577a559c520655d814cd8f45781dd27f29c4d/cmd/cli/common/engine_test.go#L136-L156

Changes:
- Add strict checks in layout creation to avoid accidental file removal
  - Ensure that absolute path in tmp, not literal path
  - Remove only if target is a symlink
- Split symlink cleanup and creation to multiple functions with separate tests
- Move symlink handling to utils package

### DIFF
--- a/cmd/cli/common/engine.go
+++ b/cmd/cli/common/engine.go
@@ -11,6 +11,7 @@ import (
 	"github.com/canonical/inference-snaps-cli/pkg/hardware_info"
 	"github.com/canonical/inference-snaps-cli/pkg/selector"
 	"github.com/canonical/inference-snaps-cli/pkg/storage"
+	"github.com/canonical/inference-snaps-cli/pkg/utils"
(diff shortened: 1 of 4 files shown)


farshidtz:
Rewrote commits to add signatures.

mrmara (approved):

PR #323, opened 2026-04-13
Add `webui` command to launch web UI in browser

jpm-canonical:
### Add `webui` command

To enable the web UI command in the CLI, one needs to add it to the additional features env var:
```
apps:
  cli:
    ...
    plugs:
      - desktop # for xdg-open
      - ...
    environment:
      ADDITIONAL_FEATURES: chat, webui
```

Install snap, run `webui` subcommand. After pressing Enter a browser tab opens at correct URL. Quitting with CTRL+C will exit without opening a browser tab.
```
$ stack-utils webui
Press Enter to open http://localhost:8081/ in the default browser ...

```

Stop openai server and try again. Browser does not launch:
```
$ sudo snap stop stack-utils.server 
Stopped.
$ stack-utils webui
Error: server not active

Run "sudo snap start stack-utils.server" to start it.
$ sudo snap start stack-utils.server 
Started.
$ stack-utils webui
Press Enter to open http://localhost:8081/ in the default browser ...
^C
```

Stop UI server and try again. Browser does not launch:
```
$ sudo snap stop stack-utils.server-webui 
Stopped.
$ stack-utils webui
Error: server-webui not active

Run "sudo snap start stack-utils.server-webui" to start it.
$ sudo snap start stack-utils.server-webui 
Started.
$ stack-utils webui
Press Enter to open http://localhost:8081/ in the default browser ...
^C
$ 
```

The `webui` command waits for the `openai` endpoint to be ready, just like the `chat` command. The same spinners that are used by the chat command are reused here.

### Status reports webui endpoint

When the `webui` feature is added to the ADDITION_FEATURES env var, the webui endpoint will be added to the status endpoints.

```
$ stack-utils status
engine: intel-cpu
services:
    server: active
    ui-server: active
endpoints:
    kserve: http://localhost:8080/2
    openai: http://localhost:8080/v3
    openvino: http://localhost:8080/v1
    webui: http://localhost:8081/
```

### Mock OpenAI server

A python script is added to `test_data` which serves the `/models` and `/chat/completions` endpoints. It was written using Copilot. 

This script is run when the server service in the `stack-utils` snap is started. This is useful for testing the chat and webui with predictable responses. A 0.5s (configurable) delay is added between responses to make the wait spinners of `chat` and `webui`, as well as UI progress more visible.


cmd/cli/main.go
jpm-canonical, line 104: It feels like `serve-ui` should also be guarded by the feature flag. If we do that, we also need to add the additional features env var to the `ui-server` app, otherwise the serve-ui script won't have access to this command.

farshidtz: I think this is okay as-is because `serve-ui` command is hidden and an app to use it is already optional.

cmd/cli/commands/ui.go
```diff
@@ -0,0 +1,91 @@
...
+	// Check ui server and engine server
+	checkServices := []string{"ui-server", "server"}
+	for _, service := range checkServices {
+		uiServerStatus, ok := services[service]
+		if !ok {
+			return fmt.Errorf("%s: service not found", service)
+		}
+		if uiServerStatus == "inactive" {
+			return fmt.Errorf("%s.%s not active\n\n%s",
+				env.SnapInstanceName(), service, common.SuggestStartServer())
```
jpm-canonical: `SuggestStartServer` prints `Run "sudo snap start stack-utils.server" to start the server.`. This is incorrect if the `ui-server` is the one that is not running. We need to replace this error with an alternative. Perhaps simply:

```
$ stack-utils ui
Error: stack-utils.ui-server not active

Start the server and try again.
```

farshidtz: I see there is a loop that checks both services. How about adding a SuggestStartService(string) which prints the suggestion depending on the input?

jpm-canonical: Much nicer:

```
$ stack-utils ui
Press Enter to open http://localhost:8081/ in the default browser ...
^C
$
$ sudo snap stop stack-utils.server-webui 
Stopped.
$
$ stack-utils ui
Error: server-webui not active

Run "sudo snap start stack-utils.server-webui" to start it.
$
$ sudo snap start stack-utils.server-webui
Started.
$
$ stack-utils ui
Press Enter to open http://localhost:8081/ in the default browser ...
^C
$
$ sudo snap stop stack-utils.server
Stopped.
$
$ stack-utils ui
Error: server not active

Run "sudo snap start stack-utils.server" to start it.
```

cmd/cli/commands/webui.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	chatClient := common.ChatClient(chatBaseUrl, "", cmd.Verbose)
+	err = chatClient.WaitChatServerReady() // prints same spinner as used for chat
+	if err != nil {
+		return err
+	}
+
+	// Print url and ask confirmation before opening
+	fmt.Printf("Press Enter to open %s in the default browser ...\n", url)
```
farshidtz, line 75: There will an extra line break with the Enter key press. How about adding line breaks before errors and:
```suggestion
	fmt.Printf("Press Enter to open %s in the default browser ...", url)
```

farshidtz (approved):
Tested with gemma3. Thanks for additional tests and refactoring!

PR #325, opened 2026-04-15
Default Y prompt, remove extra newline

farshidtz:
Change confirmation to take "Y" by default:
```console
$ sudo gemma3 use-engine cpu-xsmall 
Need to install the following components:
- llamacpp (31.9MiB)
- model-270m-it-q4-0-gguf (225.1MiB)

Do you want to continue? [Y/n] Y

Installed llamacpp
Installed model-270m-it-q4-0-gguf
Engine changed to "cpu-xsmall".

Run "snap restart gemma3.server" to use the new engine.
```

Move "Enter" prompt handling to package. Remove extra newline after the prompt message. Change prompt message.

```console
$ gemma3 webui 
Open http://localhost:8081/ in the default browser [Enter]
```

### DIFF
--- a/cmd/cli/common/prompt.go
+++ b/cmd/cli/common/prompt.go
@@ -39,20 +39,12 @@ func ConfirmationPrompt(prompt string) bool {
 func ConfirmationPromptEnter(prompt string) bool {
 	reader := bufio.NewReader(os.Stdin)
 
-	for {
-		fmt.Printf("%s [Enter] ", prompt)
+	fmt.Printf("%s Press [Enter] to continue... ", prompt)
 
-		input, err := reader.ReadString('\n')
-		if err != nil {
-			fmt.Printf("Error reading input: %v\n", err)
-			continue
-		}
-		input = strings.TrimSpace(input)
-
-		if input == "" {
-			return true
-		} else {
-			fmt.Println(`Invalid input. Please press "Enter" to continue or "Ctrl+C" to cancel.`)
-		}
+	_, err := reader.ReadString('\n')
+	if err != nil {
+		fmt.Printf("\nError reading input: %v\n", err)
+		return false
 	}
+	return true
 }
--- a/cmd/cli/common/prompt.go
+++ b/cmd/cli/common/prompt.go
@@ -7,13 +7,16 @@ import (
 	"strings"
 )
 
-// ConfirmationPrompt prompts the user and returns true for 'y', false for 'n'.
-// It defaults to true if the user presses Enter without a response.
-func ConfirmationPrompt(prompt string) bool {
+// Prompt prompts the user and returns true for 'y', false for 'n'.
+func PromptYN(prompt string, defaultResponse bool) bool {
 	reader := bufio.NewReader(os.Stdin)
 
 	for {
-		fmt.Printf("%s [Y/n] ", prompt) // default to yes
+		if defaultResponse == true {
+			fmt.Printf("%s [Y/n] ", prompt) // default is yes
+		} else {
+			fmt.Printf("%s [y/N] ", prompt) // default is no
+		}
 
 		input, err := reader.ReadString('\n')
 		if err != nil {
@@ -23,8 +26,8 @@ func ConfirmationPrompt(prompt string) bool {
 
 		input = strings.ToLower(strings.TrimSpace(input))
 		switch input {
-		case "": // default to yes on empty input
-			return true
+		case "": // default on empty input
+			return defaultResponse
 		case "Y", "y":
 			return true
 		case "N", "n":
--- a/cmd/cli/common/prompt.go
+++ b/cmd/cli/common/prompt.go
@@ -38,16 +38,21 @@ func PromptYN(prompt string, defaultResponse bool) bool {
 	}
 }
 
-// ConfirmationPromptEnter prompts the user for Enter
-func ConfirmationPromptEnter(prompt string) bool {
+// PromptEnterln prompts the user for Enter in a new line
+func PromptlnEnter(action string) bool {
 	reader := bufio.NewReader(os.Stdin)
 
-	fmt.Printf("%s Press [Enter] to continue... ", prompt)
+	fmt.Printf("Press [Enter] to %s, or [q] to abort. ", action)
 
-	_, err := reader.ReadString('\n')
+	input, err := reader.ReadString('\n')
 	if err != nil {
 		fmt.Printf("\nError reading input: %v\n", err)
 		return false
 	}
+
+	if strings.TrimSpace(input) == "q" {
+		return false
+	}
+
 	return true
 }
--- a/cmd/cli/common/prompt.go
+++ b/cmd/cli/common/prompt.go
@@ -38,19 +38,15 @@ func PromptYN(prompt string, defaultResponse bool) bool {
 	}
 }
 
-// PromptEnterln prompts the user for Enter in a new line
+// PromptlnEnter prompts the user for Enter in a new line
 func PromptlnEnter(action string) bool {
 	reader := bufio.NewReader(os.Stdin)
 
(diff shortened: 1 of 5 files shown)


jpm-canonical:
Y/n prompt works great:
```
$ sudo stack-utils use-engine intel-cpu
[sudo: authenticate] Password:        
Need to install the following components:
- dummy-component-1

Do you want to continue? [Y/n] 

Error: installing missing components: installing components: snap not known to the store:
Rerun this command after manually installing "dummy-component-1"
$
$ sudo stack-utils use-engine intel-cpu
Need to install the following components:
- dummy-component-1

Do you want to continue? [Y/n] n
Cancelled. No changes applied.
$
$ sudo stack-utils use-engine intel-cpu
Need to install the following components:
- dummy-component-1

Do you want to continue? [Y/n] d
Invalid input. Please enter "y" or "n".
Do you want to continue? [Y/n] ^C

```

Enter prompt works, but I'm not sure about the UX:
```
$ stack-utils webui 
Open http://localhost:8081/ in the default browser [Enter] a
Invalid input. Please press "Enter" to continue or "Ctrl+C" to cancel.
Open http://localhost:8081/ in the default browser [Enter] s
Invalid input. Please press "Enter" to continue or "Ctrl+C" to cancel.
Open http://localhost:8081/ in the default browser [Enter] aaasadsf
Invalid input. Please press "Enter" to continue or "Ctrl+C" to cancel.
Open http://localhost:8081/ in the default browser [Enter] ^C
$
$ stack-utils webui 
Open http://localhost:8081/ in the default browser [Enter] 
```
It feels a little weird that I can type text after the prompt. And then when I press enter, I am told it's invalid and I have to press enter. How about using the same [Y/n] prompt here too, as the behavior will be the same: pressing Enter opens the browser.

farshidtz:
> It feels a little weird that I can type text after the prompt. And then when I press enter, I am told it's invalid and I have to press enter. How about using the same [Y/n] prompt here too, as the behavior will be the same: pressing Enter opens the browser.

Opening the web ui is the active intent if this command. A question usually follows prior explanations, e.g.:
```
$ gemma3 webui 
Web UI is available at http://localhost:8081/
Open it in the default browser? [Y/n] 
```

I am also not sure if every user knows that a capital Y means that you can just press enter for a positive response and vice versa with N.

Why are you inputting text and expecting it not to fail? It says the only accepted input is Enter.

jpm-canonical:
> Why are you inputting text and expecting it not to fail?

I'm expecting it to fail the moment I enter the invalid character (press the invalid key). Not a while later when I press enter.

hartmutobendorf:
Hi, I was added to the conversation. I believe what JP is expecting is sensible – for an algorithmophile. It follows the convention that direct feedback is better than delayed feedback.

I guess a more standard behaviour would be to ignore all input except the triggers for continue and abort. But that would require you to also mention the option of _not_ opening the browser window.

`Opening http://localhost:8081/ in the default browser. Press [Enter] to continue, or [q] to abort.`

farshidtz:
Thanks for the feedback. I've reverted the input handling to its original state and extended the message to:
```console
$ gemma3 webui 
Opening http://localhost:8081/ in the default browser. Press [Enter] to continue ... ^C
$ gemma3 webui 
Opening http://localhost:8081/ in the default browser. Press [Enter] to continue ... 
```
I went with a shorter version, without `or [q] to abort`, because the extra text and option make it hard to scan through the message quickly. 

Here is a similar reference:
```console
$ gh auth login
? What account do you want to log into? GitHub.com
? What is your preferred protocol for Git operations on this host? HTTPS
? How would you like to authenticate GitHub CLI? Login with a web browser

! First copy your one-time code: xxxx-xxxx
Press Enter to open github.com in your browser... abc   
Opening in existing browser session.
^C
```

jpm-canonical (approved):
This UX is better. 

Just a pity we are wrapping over the 80 chars now.
Image

hartmutobendorf:
Perhaps you can add a \n before Press...

farshidtz:
It can be made shorter to fit under 80 chars:
```
$ stack-utils webui 
Opening http://localhost:8081/ in your browser. Press [Enter] to continue... ^C
```

@hartmutobendorf  With a line break, it feel misleading to say "Opening" but not doing it yet. How about the following, which provides a clearer path to let the user decide what to do with the URL, i.e. either Ctrl+Click to open it or press Enter:
```
$ stack-utils webui  
Web UI is available at http://localhost:8081/
Press [Enter] to open it in your browser...
```

hartmutobendorf:
I like that better because it allows you to describe the alternative ([q]).

farshidtz:
Changed to:
```console
$ stack-utils webui 
Web UI is available at http://localhost:8081/
Press [Enter] to open it in your browser, or [q] to abort.
```

Changed prune-cache's default to N / false.

Added tests because this is getting hard to manually validate.

cmd/cli/common/prompt.go
```diff
@@ -7,21 +7,52 @@ import (
...
+		case "N", "n":
 			return false
-		} else {
+		default:
 			fmt.Println(`Invalid input. Please enter "y" or "n".`)
 		}
 	}
 }
+
+// PromptEnterln prompts the user for Enter in a new line
```
jpm-canonical:
```suggestion
// PromptlnEnter prompts the user for Enter in a new line
```

cmd/cli/common/prompt.go
```diff
@@ -7,21 +7,52 @@ import (
...
-		fmt.Printf("%s [y/n] ", prompt)
-		input, _ := reader.ReadString('\n')
-		input = strings.ToLower(strings.TrimSpace(input))
+		if defaultResponse == true {
+			fmt.Printf("%s [Y/n] ", prompt) // default is yes
+		} else {
+			fmt.Printf("%s [y/N] ", prompt) // default is no
+		}
+
+		input, err := reader.ReadString('\n')
```
jpm-canonical, line 21: I was thinking about using `reader.ReadRune()` here. Then the first keystroke would make it error out, rather than printing bogus characters to stdout.

jpm-canonical (approved):
I think this is ok. My preference would be to avoid printing invalid inputs chars to the terminal in the Enter case, but we've seen `gh cli` does this too.

```
$ sudo stack-utils use-engine intel-cpu
[sudo: authenticate] Password:        
Need to install the following components:
- dummy-component-1

Do you want to continue? [Y/n] 

Error: installing missing components: installing components: snap not known to the store:
Rerun this command after manually installing "dummy-component-1"
```

```
$ sudo snap start stack-utils.server-webui 
Started.
$ stack-utils webui 
Web UI is available at http://localhost:8081/
Press [Enter] to open it in your browser, or [q] to abort. q
$ stack-utils webui 
Web UI is available at http://localhost:8081/
Press [Enter] to open it in your browser, or [q] to abort.  sdasdf^C
$ stack-utils webui 
Web UI is available at http://localhost:8081/
Press [Enter] to open it in your browser, or [q] to abort. q
$ stack-utils webui 
Web UI is available at http://localhost:8081/
Press [Enter] to open it in your browser, or [q] to abort. 
$
```

cmd/cli/common/prompt.go
farshidtz: Apologies, I do not understand. The bogus character that you type in yourself?

This is normal in most CLIs.

Here is apt:
```
Continue? [Y/n] ddsdd
Abort.
```

farshidtz:
I replaced [q] with [Ctrl+C] because this is the universal standard for aborting a foreground process. I found [q] mostly used in TUIs, and proper handling of it in a CLI (capturing without Enter) requires additional implementation that goes beyond the original purpose of this PR.

```
$ stack-utils webui 
Web UI is available at http://localhost:8081/
Press [Enter] to open it in your browser, or [Ctrl+C] to abort. ^C
```

Thanks again for your thorough testing and feedback.

PR #328, opened 2026-04-16
Implement environment variable passthrough

mrmara:
This implementation adds the possibility to override internal snap configurations. 
The `passthrough` special keyword is used to bypass check on setting only known configuration fields.

This PR covers only environment variables overriding, usage e.g.:

 ```
 gemma3 set passthrough.environment.kebab-case-envvar-name={VALUE}
 ```
 
 Environment variable name will be converted to `snake_case` and `-` will be substituted with `_`, eg:
 
 ```
 gemma3 set passthrough.environment.model-file=/home/models/gemma-3-12b-it-q4_0.gguf
 ```
 ENVVAR: `MODEL_FILE`
 VALUE: `/home/models/gemma-3-12b-it-q4_0.gguf`


cmd/cli/common/component.go
```diff
@@ -91,3 +91,34 @@ func checkMissingComponents(manifest *engines.Manifest) ([]string, error) {
 
 	return missing, nil
 }
+
+func ProcessPassthroughConfigs(ctx *Context) error {
```
farshidtz: If there is nothing common about this, keep it alongside the command itself.

farshidtz: How about having this function only extract and return the environment key values and doing the env setting separately?

cmd/cli/commands/run.go
```diff
@@ -50,6 +50,11 @@ func (cmd *runCommand) run(_ *cobra.Command, args []string) error {
 		return fmt.Errorf("loading engine environment: %v", err)
 	}
 
+	err = common.ProcessPassthroughConfigs(cmd.Context)
+	if err != nil {
+		// Print warning but continue with execution, as passthrough configs are optional and should not block the execution of the subprocess.
+		fmt.Fprintf(os.Stderr, "Warning: processing passthrough configs: %v\n", err)
+	}
```
farshidtz: I would make this a fatal error. It is better to fail early than have the process run in an unexpected way and complicate debugging.

cmd/cli/common/component.go
mrmara: In that case I would need to call two different methods in the `run` method, it may be better to separate the implementation in 3 methods:

- extractPassthroughConfigs
- setEnvironmentVariables (that can be implemented in utils)
- processPassthroughConfigs (extractPassthroughConfigs extract and setEnvironmentVariables)

farshidtz: It is up to you, keeping in mind modularity and testability.

cmd/cli/commands/run.go
```diff
@@ -61,3 +66,51 @@ func (cmd *runCommand) run(_ *cobra.Command, args []string) error {
...
+	return passthroughConfigs, nil
+}
+
+func getEnvVarsFromPassthroughConfigs(envVars map[string]any) (map[string]any, error) {
+	result := make(map[string]any)
+	for k, v := range envVars {
+		if strings.HasPrefix(k, "environment.") {
+			envVarName := strings.TrimPrefix(k, "environment.")
+			envVarValue := fmt.Sprintf("%v", v)
+			// envVarName is kebab-case and lower case, environment variables conventionally are upper case and snake case, so convert kebab-case to snake_case and uppercase the env var name
```
farshidtz:
```suggestion
			// Convert passthrough keys (kebab-case) to environment variables names (SNAKE_CASE)
```

Or rather, not getting into the casing conventions may provide false expectations:
```suggestion
			// Convert passthrough keys (my-key) to environment variables names (MY_KEY)
```

cmd/cli/commands/run.go
```diff
@@ -50,6 +52,9 @@ func (cmd *runCommand) run(_ *cobra.Command, args []string) error {
 		return fmt.Errorf("loading engine environment: %v", err)
 	}
 
+	if err := processPassthroughConfigs(cmd.Context); err != nil {
+		return fmt.Errorf("processing passthrough configs: %v", err)
+	}
```
farshidtz: This is inserted between LoadEngineEnvironment and deferring it's cleanup. An error in processPassthroughConfigs, will return without doing the cleanup.

cmd/cli/commands/run.go
```diff
@@ -61,3 +66,51 @@ func (cmd *runCommand) run(_ *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+	envVars, err := getEnvVarsFromPassthroughConfigs(passthroughConfigs)
+	if err != nil {
+		return err
+	}
+	err = utils.SetEnvironmentVariables(envVars)
+	if err != nil {
+		return fmt.Errorf("setting environment variables from passthrough configs: %v", err)
```
farshidtz: The parent already adds the context. 
```suggestion
		return fmt.Errorf("setting environment variables: %v", err)
```

Resulting in error: `processing passthrough configs: setting environment variables: <err>`

cmd/cli/commands/run.go
farshidtz, line 1: Would be great to have some tests for the new functions.

farshidtz (changes requested):
That works like a charm!

I've tested it with gemma3:
```
2026-04-17T15:49:12+02:00 gemma3.server[129424]: + exec llama-server --model /home/farshid/models/gemma-3-12b-it-q4_0.gguf --port 8328 --host 127.0.0.1 --no-warmup --mmproj /home/farshid/models/mmproj-model-f16-12B.gguf --sleep-idle-seconds 600
```

This was after setting:
```shell
sudo snap connect gemma3:home
sudo gemma3 set passthrough.environment.mmproj-file=/home/farshid/models/mmproj-model-f16-12B.gguf 
sudo gemma3 set passthrough.environment.model-file=/home/farshid/models/gemma-3-12b-it-q4_0.gguf 
```

cmd/cli/commands/run.go
mrmara: Added run_test.go. I slightly changed one method to increase coverage

cmd/cli/commands/run.go
```diff
@@ -54,10 +56,61 @@ func (cmd *runCommand) run(_ *cobra.Command, args []string) error {
...
+			envVarName = strings.ToUpper(strings.ReplaceAll(envVarName, "-", "_"))
+			result[envVarName] = envVarValue
+		}
+	}
+
+	return result, nil
+}
+
+func processPassthroughConfigs(ctx *common.Context) error {
+	configs, err := ctx.Config.GetAll()
```
farshidtz: This could just query what it needs, using the Get method.

farshidtz: You can add tests `processPassthroughConfigs` by mocking storage (see for example: https://github.com/canonical/inference-snaps-cli/blob/13e57ef513997fdc7f6e1ae1b2ee40a86a2176f2/cmd/cli/common/endpoints_test.go#L73-L74)

cmd/cli/commands/run.go
```diff
@@ -54,10 +56,61 @@ func (cmd *runCommand) run(_ *cobra.Command, args []string) error {
...
+
 	path := args[0]
 
 	execCmd := exec.Command(path)
 	execCmd.Stdout = os.Stdout
 	execCmd.Stderr = os.Stderr
 	return execCmd.Run()
 }
+
+func extractPassthroughConfigs(configs map[string]any) (map[string]any, error) {
```
farshidtz: I just noticed that the added functions are not methods of runCommand. This is needed for better encapsulation since the commands package includes many other things.

farshidtz (changes requested):
Thanks for adding the tests. I think there is still room for extending the coverage. I also suggested a couple more improvements.

farshidtz (approved):

PR #331, opened 2026-04-20
Restart services after setting configuration

### LINKED ISSUE #164: Snap on unusable state when configuration is changed and engined switched with `use-engine`
I've had the situation where this configuration was changed. But the use-engine test failed. This left the snap on my host system in an unusable state. I had to manually restart the service.

To make things more deterministic, let's rather revert the port to the original value at the end of this sub-test. Then, if the following sub-tests fail, the host system is not effected.

_Originally posted by @jpm-canonical in https://github.com/canonical/famous-models-dev/pull/7#discussion_r2405686081_

farshidtz:
Closes https://github.com/canonical/inference-snaps/issues/164

```console
$ sudo stack-utils set x=1
Error: unknown key: "x"
$ sudo stack-utils set verbose=1
Apply changes and restart stack-utils? [Y/n] n
Cancelled. Discarded configurations.
$ sudo stack-utils set verbose=1
Apply changes and restart stack-utils? [Y/n] y
$ sudo stack-utils set verbose=1 # same value ignored
$ sudo stack-utils set verbose=2 --no-restart 
Run "snap restart stack-utils" to apply the changes.
```


cmd/cli/commands/set.go
```diff
@@ -70,12 +73,43 @@ func (cmd *setCommand) setValue(keyValue string) error {
...
+		}
+
+		if fmt.Sprint(currVal) == value {
+			return nil // no change needed
+		}
+
+		if !cmd.noRestart {
+			msg := fmt.Sprintf("Apply changes and restart %s?", cmd.Snap.InstanceName())
+			if !(cmd.assumeYes || common.PromptYN(msg, true)) {
+				fmt.Println("Cancelled. Discarded configurations.")
```
jpm-canonical: "Discarded configurations." was a bit unexpected when I read it the first time. I interpreted it as if all previous configurations of the snap was discarded.

How about we make it a little more explicit, using the same terminology as the line printed above it?
```suggestion
				fmt.Println("Cancelled. Changes not applied.")
```

pkg/snap/snap.go
```diff
@@ -0,0 +1,31 @@
...
+type Snap interface {
+	Restart(service ...string) error
+	InstanceName() string
+}
+
+func New() Snap {
+	return &snap{}
+}
+
+type snap struct{}
```
farshidtz, line 17: This could eventually hold the connection for https://github.com/canonical/go-snapctl/issues/9

jpm-canonical (approved):
Looks good. Maybe we can reconsider the negative case's log message.

PR #332, opened 2026-04-20
Enable passing program arguments to run

mrmara:
This implementation adds the possibility to pass to `run` command an entire command to run.
Sperator `--` is exploited to use the feature.

E.G.

`sudo SNAP_NAME run -- echo "Hello World!"`


cmd/cli/commands/run.go
```diff
@@ -23,10 +23,10 @@ func Run(ctx *common.Context) *cobra.Command {
 	cmd.Context = ctx
 
 	cobraCmd := &cobra.Command{
-		Use:               "run <path>",
+		Use:               "run <path> [args...]",
```
farshidtz: This results in wrong usage:
```console
$ go run ./cmd/cli run -h
Run a subprocess

Usage:
  cli run <path> [flags]

Flags:
  -h, --help                  help for run
      --wait-for-components   wait for engine components to be installed before running

Global Flags:
  -v, --verbose   Enable verbose logging
```

Suggesting e.g. `cli run ls -l --wait-for-components` instead of `cli run --wait-for-components -- ls -l`.

"Use" can stay simple, e.g.:
```
Usage:
  cli run <command> [flags]
```

And extended usage can go to a long description : ` cli run [flags] -- <command> [command's args...]`.

farshidtz (changes requested):
This works as expected. I really like the simple implementation and use of double dashes to treat everything that follows as positional arguments.

I've some suggestions regarding the usage documentation.

farshidtz (approved):

PR #333, opened 2026-04-21
Make restart optional on set

farshidtz:
The mandatory restart implemented by #331 is awkward in practice, as it prevents command chaining without restarts when the --no-restart is not set. This was by design, nevertheless, it does not feel natural to revert configurations (the primary intent of the command) just because a restart isn't wanted at that point. I ran into it when trying to replicate the steps given [here](https://github.com/canonical/inference-snaps-cli/pull/328#pullrequestreview-4128964176).

I've also done some refactoring in the set command for better testability. The returned errors from the two new methods are passed as is (no context added to the error chains). I consider this a reasonable deviation for when calling sibling methods.

```console
$ stack-utils get
http.host: 127.0.0.1
http.port: 8322
verbose: 1
webui.http.host: 127.0.0.1
webui.http.port: 8323
$ stack-utils set x=1
Error: permission denied, try again with sudo
$ sudo stack-utils set x=1
Error: key "x" is not found
$ sudo stack-utils set verbose=1
$ sudo stack-utils set verbose=2
Restart stack-utils to apply the changes? [Y/n] n
$ sudo stack-utils set verbose=2
$ sudo stack-utils set verbose=3
Restart stack-utils to apply the changes? [Y/n] y
```

### DIFF
--- a/cmd/cli/commands/set.go
+++ b/cmd/cli/commands/set.go
@@ -57,59 +57,69 @@ func (cmd *setCommand) run(_ *cobra.Command, args []string) error {
 }
 
 func (cmd *setCommand) setValue(keyValue string) error {
+	key, value, err := cmd.parseKeyValue(keyValue)
+	if err != nil {
+		return err
+	}
+
+	if cmd.packageConfig {
+		if err := cmd.Config.Set(key, value, storage.PackageConfig); err != nil {
+			return fmt.Errorf("setting %q to %q: %v", key, value, err)
+		}
+	} else if cmd.engineConfig {
+		if err := cmd.Config.Set(key, value, storage.EngineConfig); err != nil {
+			return fmt.Errorf("setting %q to %q: %v", key, value, err)
+		}
+	}
+
+	return cmd.setUserConfig(key, value)
+}
+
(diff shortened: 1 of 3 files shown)


cmd/cli/commands/set.go
```diff
@@ -57,59 +57,69 @@ func (cmd *setCommand) run(_ *cobra.Command, args []string) error {
...
+		if err := cmd.Config.Set(key, value, storage.EngineConfig); err != nil {
+			return fmt.Errorf("setting %q to %q: %v", key, value, err)
+		}
+	}
+
+	return cmd.setUserConfig(key, value)
+}
+
+func (cmd *setCommand) parseKeyValue(keyValue string) (key, value string, err error) {
+	if keyValue == "" {
```
jpm-canonical, line 79: Can't an empty string be a valid config value?

jpm-canonical (approved):

cmd/cli/commands/set.go
farshidtz: Yes, it could be, for string values. We can adjust this separately. This PR is only refactoring the key-value validation and not introducing this restriction.

PR #335, opened 2026-04-21
Add Qualcomm Dragonwing NPU detection via FastRPC and selector support

Sinan-Karakaya:
This PR adds proper support for Qualcomm Dragonwing NPUs that show up through FastRPC device nodes.

Until now, our detection/selection flow was mostly centered around PCI/USB-style signals. On Dragonwing systems that meant we could miss valid NPU-capable setups, or fail to confidently match engines that should run there. This change closes that gap.

# What’s in here
Added new hardware info fields for detected platforms/devices.
Added Qualcomm detection logic that looks at FastRPC nodes (`/dev/fastrpc-*`, especially CDSP nodes).
Added `fastrpc` as a supported bus in engine/device validation.
Added selector matching/scoring logic for FastRPC devices.
Added tests across detection, validation, and selector behavior.
Small robustness fix: disk info collection now skips missing directories instead of failing hard.

# Why this matters
On Dragonwing hardware, NPU availability is represented differently from traditional PCI GPU-style setups. By recognizing FastRPC directly, we can now:
- report the platform/device state more accurately, and
- successfully select engines that declare `type: npu` + `bus: fastrpc`.

# Behavior after this PR
On a Dragonwing machine with FastRPC CDSP nodes present:
- `show-machine` includes Qualcomm/Dragonwing in platforms and an NPU FastRPC device entry.
- `debug select-engine` can positively match engines that target FastRPC NPUs.

# Compatibility
This is additive. Existing PCI/USB paths are unchanged, and current manifests continue to behave as before.


pkg/types/hw_info.go
```diff
@@ -5,4 +5,6 @@ type HwInfo struct {
 	Memory     MemoryInfo          `json:"memory,omitempty" yaml:"memory,omitempty"`
 	Disk       map[string]DirStats `json:"disk,omitempty" yaml:"disk,omitempty"`
 	PciDevices []PciDevice         `json:"pci,omitempty" yaml:"pci"`
+	Platforms  []PlatformInfo      `json:"platforms,omitempty" yaml:"platforms,omitempty"`
+	Devices    []DetectedDevice    `json:"devices,omitempty" yaml:"devices,omitempty"`
```
talhaHavadar: I have concerns, that we are going to end up in a messy struct if we go this way.

Pci is a bus, PciDevice is a device on pci bus

Devices is like all devices? (excluding the ones on pci bus??)
And I see in DetectedDevice struct we have Bus field but then why pcidevices can not be part of Devices?

This will require some fundamental changes in this PR so feel free to challenge or ask questions if you need things to be cleared

talhaHavadar (changes requested):

pkg/types/hw_info.go
Sinan-Karakaya: Mmh right, the option I could see to fix that would be to consolidate everything under `DetectedDevice` and maybe deprecate `PciDevices`, but that's not ideal either, I don't want to introduce breaking changes.

I thought about maybe merging the 2 kind of devices together, but they map very differently (`PciDevice` is a detailed catalog entry from `lspci`, while `DetectedDevice` is a runtime signal that a device exists, regardless of its bus).

So I'm not sure if there would have been a cleaner solution than I could have done. I can document a bit more `DetectedDevice` to emphasize that it serves a different purpose than `PciDevice`

talhaHavadar: I would prefer to have cleaner solution even if it causes breaking changes as it maybe way harder to change things later once they are grown so much. (Thats why versioning exist I think)

But that said I am not the owner of this repo/project so it is up to @farshidtz and/or @jpm-canonical decide if they dont want breaking changes.

Sinan-Karakaya: I already made @jpm-canonical aware of this issue, they'll discuss it in their team and come back to us for this issue :)

pkg/types/detected_devices.go
```diff
@@ -0,0 +1,13 @@
+package types
+
+type PlatformInfo struct {
+	Vendor string `json:"vendor" yaml:"vendor"`
+	Name   string `json:"name" yaml:"name"`
+	SoC  string `json:"soc,omitempty" yaml:"soc,omitempty"`
```
talhaHavadar: I think SoC is too specific, can we just have VendorName, ProductName or just Vendor, Product

And instead of PlatformInfo as struct name, we can have DeviceMetadata as struct name

Sinan-Karakaya: Done, renamed PlatformInfo->DeviceMetadata and SoC->VendorName

pkg/types/detected_devices.go
```diff
@@ -0,0 +1,13 @@
+package types
+
+type DeviceMetadata struct {
+	Vendor     string `json:"vendor" yaml:"vendor"`
+	Name       string `json:"name" yaml:"name"`
+	VendorName string `json:"vendor_name,omitempty" yaml:"vendor_name,omitempty"`
```
talhaHavadar: I meant SoC could be ProductName so we end up, VendorName, ProductName in DeviceMetadata struct therefore "Name" would not be needed anymore.

Sinan-Karakaya: Aaah I see, my bad that's a quick fix!

farshidtz:
Corresponding PR for the detection moved to lscompute: https://github.com/canonical/lscompute/pull/15

jpm-canonical:
The hardware detection part of this change has been merged into lscompute in [this PR](https://github.com/canonical/lscompute/pull/15). And this project has been updated to use lscompute for hardware detection. The selection logic from this PR is however still missing. @Sinan-Karakaya could you refactor this PR, or create a new one, that adds the selection logic for fastrpc devices to this project?

Sinan-Karakaya:
@jpm-canonical should be ready for review

pkg/selector/fastrpc/fastrpc.go
```diff
@@ -0,0 +1,50 @@
...
+
+		return weights.PciDevice + weights.PciDeviceType, nil
+	}
+
+	return 0, []string{"no fastrpc devices on host system"}
+}
+
+func checkSnapConnection(connection string) (bool, error) {
+	if testing.Testing() {
+		// Tests do not necessarily run inside a snap.
```
jpm-canonical: To make testing easier, we can check that we're inside a snap before checking snap connections. 
```suggestion
	if os.Getenv("SNAP") == "" {
		// Snap connections can only be queried from inside a snap.
```

Then you can test like this:
```
$ cat qualcomm-machine-info.json | go run ./cmd/modelctl debug select-engine --engines ./test_data/engines/ 1>/dev/null
```

Or on a qualcomm device:
```
$ go run ./cmd/modelctl show-machine --format=json | go run ./cmd/modelctl debug select-engine --engines ./test_data/engines/ 1>/dev/null
```

pkg/selector/fastrpc/fastrpc_test.go
```diff
@@ -0,0 +1,53 @@
...
+			},
+		}
+		device := engines.Device{Type: "npu", Bus: "fastrpc"}
+
+		score, issues := Match(device, machineInfo)
+		if score == 0 || len(issues) > 0 {
+			t.Fatalf("expected a positive score with no issues, score=%d issues=%v", score, issues)
+		}
+	})
+
```
jpm-canonical, line 28: I suggest adding these for better testing of the logic. A type of `npu` in the engine manifest, should only match a fastrpc `cdsp` domain device. If the `type` is omitted from the engine manifest, and the bus is set to `fastrpc`, then any domain should match.

I think this is going to show an issue with the matching logic.

```suggestion

	t.Run("FastRPC NPU rejects non-CDSP domains", func(t *testing.T) {
		domains := []lsfastrpc.FastRPCDomain{
			lsfastrpc.ADSPDomain,
			lsfastrpc.MDSPDomain,
			lsfastrpc.SDSPDomain,
			lsfastrpc.GDSPDomain,
		}
		for _, domain := range domains {
			t.Run(string(domain), func(t *testing.T) {
				machineInfo := &machine.MachineInfo{
					Devices: []any{
						lsfastrpc.Device{
							Bus:    lsfastrpc.BusName,
							Domain: domain,
						},
					},
				}
				device := engines.Device{Type: "npu", Bus: "fastrpc"}

				score, issues := Match(device, machineInfo)
				if score != 0 {
					t.Fatalf("expected score=0, got %d", score)
				}
				if len(issues) == 0 {
					t.Fatal("expected compatibility issues")
				}
			})
		}
	})

	t.Run("FastRPC without device type accepts any domain", func(t *testing.T) {
		domains := []lsfastrpc.FastRPCDomain{
			lsfastrpc.ADSPDomain,
			lsfastrpc.MDSPDomain,
			lsfastrpc.SDSPDomain,
			lsfastrpc.CDSPDomain,
			lsfastrpc.GDSPDomain,
		}
		for _, domain := range domains {
			t.Run(string(domain), func(t *testing.T) {
				machineInfo := &machine.MachineInfo{
					Devices: []any{
						lsfastrpc.Device{
							Bus:    lsfastrpc.BusName,
							Domain: domain,
						},
					},
				}
				device := engines.Device{Bus: "fastrpc"}

				score, issues := Match(device, machineInfo)
				if score == 0 || len(issues) > 0 {
					t.Fatalf("expected a positive score with no issues, score=%d issues=%v", score, issues)
				}
			})
		}
	})
	
```

pkg/selector/fastrpc_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"testing"
+
+	"github.com/canonical/inference-snaps-cli/v2/pkg/engines"
+	"github.com/canonical/lscompute/pkg/machine"
+	lsfastrpc "github.com/canonical/lscompute/pkg/machine/device/fastrpc"
+)
+
+func TestFastRPCNPUSelection(t *testing.T) {
```
jpm-canonical, line 11: Maybe add a test for the negative case here too.

pkg/types/detected_devices.go
```diff
@@ -0,0 +1,12 @@
+package types
```
jpm-canonical: These seem to be unused. Please remove.

jpm-canonical (changes requested):

pkg/selector/select_stack.go
```diff
@@ -137,25 +146,33 @@ func scoreDevicesAny(machineInfo *machine.MachineInfo, devices []engines.Device)
...
+				devices[i].CompatibilityIssues = append(device.CompatibilityIssues, deviceIssues...)
 			} else {
 				devicesFound++
 				compatibilityScore += cpuScore
 			}
 
 		} else if device.Bus == "usb" {
 			compatible = false
-			device.CompatibilityIssues = append(device.CompatibilityIssues, "usb device matching not implemented")
+			devices[i].CompatibilityIssues = append(devices[i].CompatibilityIssues, "usb device matching not implemented")
```
jpm-canonical: Be consistent. Stick to either `devices[i]` or `device`.
```suggestion
			devices[i].CompatibilityIssues = append(device.CompatibilityIssues, "usb device matching not implemented")
```

README.md
```diff
@@ -71,6 +71,26 @@ This prints a machine-readable summary of the host system.
...
+incompatible userspace ABIs use their own versioned content contract.
+
+Platform-specific device paths, firmware, userspace libraries, and DSP files
+belong in the gadget or runtime provider. The model snap should only require
+the logical plugs and select a backend such as `fastrpc` when the detected
+device and required connections are available.
+
+FastRPC engine manifests with `type: npu` match only the CDSP domain. A
+typeless manifest with `bus: fastrpc` matches any DSP domain.
+
```
jpm-canonical: I think this should be documented somewhere else. It is important, but not directly related to this PR/repo. Should this be a spec?

jpm-canonical (changes requested):
Looks good now. Just two minor things I think we should change.

jpm-canonical (approved):
I tested this on a Qualcomm RB4, and it works as expected. Thanks a lot for the contribution!

PR #338, opened 2026-04-23
Add unset command

mrmara:
Unset command can be used to remove configuration set at `user` level.
It does not affect `engine` or `package` level.
If unset is called on a configuration which has a default value, default value will be restore. Otherwise, configuration key will be deleted

### DIFF
--- a/cmd/cli/commands/get.go
+++ b/cmd/cli/commands/get.go
@@ -43,7 +43,7 @@ func (cmd *getCommand) getValue(key string) error {
 	}
 
 	if len(value) == 0 {
-		return fmt.Errorf("%s", common.SuggestKeyNotFound(key))
+		return fmt.Errorf("Key %q is not found\n\n%s", key, common.SuggestKeyNotFound(key))
 	}
 
 	if len(value) == 1 {
--- a/cmd/cli/commands/get.go
+++ b/cmd/cli/commands/get.go
@@ -43,7 +43,7 @@ func (cmd *getCommand) getValue(key string) error {
 	}
 
 	if len(value) == 0 {
-		return fmt.Errorf("Key %q is not found\n\n%s", key, common.SuggestKeyNotFound(key))
+		return fmt.Errorf("key %q is not found\n\n%s", key, common.SuggestKeyNotFound(key))
 	}
 
 	if len(value) == 1 {
--- a/cmd/cli/commands/get.go
+++ b/cmd/cli/commands/get.go
@@ -43,7 +43,7 @@ func (cmd *getCommand) getValue(key string) error {
 	}
 
 	if len(value) == 0 {
-		return fmt.Errorf("key %q is not found\n\nUse \"%s get\" to view available keys", key, cmd.Snap.InstanceName())
+		return fmt.Errorf("%s", common.SuggestKeyNotFound(key))
 	}
 
 	if len(value) == 1 {
--- a/cmd/cli/commands/get.go
+++ b/cmd/cli/commands/get.go
@@ -43,7 +43,7 @@ func (cmd *getCommand) getValue(key string) error {
 	}
 
 	if len(value) == 0 {
-		return fmt.Errorf("no value set for key %q", key)
+		return fmt.Errorf("key %q is not found\n\nUse \"%s get\" to view available keys", key, cmd.Snap.InstanceName())
 	}
 
 	if len(value) == 1 {
--- a/cmd/cli/commands/run_test.go
+++ b/cmd/cli/commands/run_test.go
@@ -36,11 +36,9 @@ func TestGetEnvVarsFromPassthroughConfigs(t *testing.T) {
 
 func TestProcessPassthroughConfigs(t *testing.T) {
 	mockConfig := storage.NewMockConfig()
-	mockConfig.SetAll(map[string]any{
-		"passthrough.environment.my-key": "value",
-		"passthrough.environment.other":  123,
-		"passthrough.not-environment":    "ignored",
-	}, storage.UserConfig)
+	mockConfig.Set("passthrough.environment.my-key", "value", storage.UserConfig)
+	mockConfig.Set("passthrough.environment.other", "123", storage.UserConfig)
+	mockConfig.Set("passthrough.not-environment", "ignored", storage.UserConfig)
 	cmd := runCommand{
 		Context: &common.Context{
 			Config: mockConfig,
--- a/cmd/cli/commands/run_test.go
+++ b/cmd/cli/commands/run_test.go
@@ -38,9 +38,9 @@ func TestProcessPassthroughConfigs(t *testing.T) {
 	cmd := runCommand{
 		Context: &common.Context{
 			Config: storage.NewMockConfig(map[string]any{
-				"passthrough.environment.my-key": "value",
-				"passthrough.environment.other":  123,
-				"passthrough.not-environment":    "ignored",
+				"user.passthrough.environment.my-key": "value",
+				"user.passthrough.environment.other":  123,
+				"user.passthrough.not-environment":    "ignored",
 			}),
 		},
 	}
--- a/cmd/cli/commands/run_test.go
+++ b/cmd/cli/commands/run_test.go
@@ -35,13 +35,15 @@ func TestGetEnvVarsFromPassthroughConfigs(t *testing.T) {
 }
 
 func TestProcessPassthroughConfigs(t *testing.T) {
+	mockConfig := storage.NewMockConfig()
+	mockConfig.SetAll(map[string]any{
+		"passthrough.environment.my-key": "value",
+		"passthrough.environment.other":  123,
+		"passthrough.not-environment":    "ignored",
+	}, storage.UserConfig)
 	cmd := runCommand{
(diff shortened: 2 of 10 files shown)


cmd/cli/commands/run_test.go
```diff
@@ -38,9 +38,9 @@ func TestProcessPassthroughConfigs(t *testing.T) {
 	cmd := runCommand{
 		Context: &common.Context{
 			Config: storage.NewMockConfig(map[string]any{
-				"passthrough.environment.my-key": "value",
-				"passthrough.environment.other":  123,
-				"passthrough.not-environment":    "ignored",
+				"user.passthrough.environment.my-key": "value",
```
farshidtz: Adding the config type as a prefix is hacky and misleading because the config type is not supposed to be part of the config key that is exposed to the user. These are internal to the library implementation.

IMO, the constructor `NewMockConfig()` should not accept raw input. Instead, the values should be set with its Set(key, value, type) method and stored internally inside the right model. A more complicated alternative is to make the constructor receive the three config types separately (three maps).

cmd/cli/commands/set.go
```diff
@@ -102,7 +102,7 @@ func (cmd *setCommand) setUserConfig(key, value string) error {
 	}
 	currVal, found := currValMap[key]
 	if !found && !strings.HasPrefix(key, "passthrough.") {
-		return fmt.Errorf("unknown key: %q", key)
+		return fmt.Errorf("key %q is not found\n\nUse \"%s get\" to view available keys", key, cmd.Snap.InstanceName())
```
farshidtz: `\"%s get\" to view available keys` is replicated in three places. The project keeps reusable suggestions in common/suggestions.go.

cmd/cli/commands/unset.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (cmd *unsetCommand) run(_ *cobra.Command, args []string) error {
+	if !utils.IsRootUser() {
+		return common.ErrPermissionDenied
+	}
+
+	return cmd.unsetValue(args[0])
+}
+
+func (cmd *unsetCommand) unsetValue(keyValue string) error {
```
farshidtz: The input is just the key.
```suggestion
func (cmd *unsetCommand) unsetValue(key string) error {
```

cmd/cli/commands/unset.go
```diff
@@ -0,0 +1,83 @@
...
+	err = cmd.Config.Unset(keyValue, storage.UserConfig)
+	if err != nil {
+		return fmt.Errorf("unsetting %q: %v", keyValue, err)
+	}
+
+	newValMap, err := cmd.Config.Get(keyValue)
+	if err != nil {
+		return fmt.Errorf("checking existing keys: %s", err)
+	}
+	newVal, _ := newValMap[keyValue]
```
farshidtz:
```suggestion
	newVal := newValMap[keyValue]
```

cmd/cli/commands/unset.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if fmt.Sprint(currVal) == fmt.Sprint(newVal) {
+		return nil // no change needed
```
farshidtz:
```suggestion
		return nil // value not changed
```

dev
farshidtz, line 1: This doesn't look right.

cmd/cli/commands/unset_test.go
```diff
@@ -0,0 +1,81 @@
...
+	values, err = config.Get("test-key")
+	if err != nil {
+		t.Fatalf("Get returned an unexpected error: %v", err)
+	}
+	if values["test-key"] != "engine-value" {
+		t.Fatalf("expected test-key to be overridden by engine config, got %#v", values["test-key"])
+	}
+}
+
+func TestUnsetInexistentKey(t *testing.T) {
```
farshidtz:
```suggestion
func TestUnsetNonexistentKey(t *testing.T) {
```

farshidtz (changes requested):
Thanks. Great to have the mock config on par with the actual implementation with layered config and flattened keys.

cmd/cli/common/suggestions.go
```diff
@@ -66,3 +66,12 @@ func SuggestEngineInfo() string {
...
 	return fmt.Sprintf("Use \"%v show-engine <engine>\" for more information about an engine.", instanceName)
 }
+
+func SuggestKeyNotFound(key string) string {
+	instanceName := env.SnapInstanceName()
+	if instanceName == "" { // not a snap
+		instanceName = "<snap-instance-name>"
+	}
+
+	return fmt.Sprintf("Key %q is not found\n\nUse \"%s get\" to view available keys", key, instanceName)
```
farshidtz: This is more than a suggestion. It is an error, followed by a suggestion. This file is used to maintain suggestions only. This function should ideally reduce its scope to the same.

pkg/storage/mock_config.go
```diff
@@ -5,30 +5,38 @@ import (
 	"strings"
 )
 
-type mockConfig struct {
+type MockConfig struct {
 	values map[string]any
 }
 
-func NewMockConfig(values map[string]any) Config {
+func NewMockConfig() *MockConfig {
```
farshidtz: The return type should not change. To purely reflect the actual implementation, the mock backend must comply with the Config interface and only expose the methods defined by the interface.

cmd/cli/commands/run_test.go
```diff
@@ -35,13 +35,15 @@ func TestGetEnvVarsFromPassthroughConfigs(t *testing.T) {
 }
 
 func TestProcessPassthroughConfigs(t *testing.T) {
+	mockConfig := storage.NewMockConfig()
+	mockConfig.SetAll(map[string]any{
+		"passthrough.environment.my-key": "value",
+		"passthrough.environment.other":  123,
+		"passthrough.not-environment":    "ignored",
+	}, storage.UserConfig)
```
farshidtz: This could use Set or SetDocument (see Config interface). I could imagine adding a function in the future to set multiple keys at once (for `model set one=1 two=2`) but that needs to be consistent with the actual config implementation. I personally find the use of Set acceptable here, without the need for a map or JSON input as with SetDocument.

cmd/cli/commands/set_test.go
```diff
@@ -68,11 +68,14 @@ func TestParseKeyValue(t *testing.T) {
 }
 
 func TestSetValueSuccessForUserConfig(t *testing.T) {
-	config := storage.NewMockConfig(map[string]any{"user.api.endpoint": "https://old.example.com"})
+	mockConfig := storage.NewMockConfig()
+	mockConfig.SetAll(map[string]any{
+		"api.endpoint": "https://old.example.com",
+	}, storage.UserConfig)
```
farshidtz:
```suggestion
	mockConfig.Set("api.endpoint", "https://old.example.com", storage.UserConfig)
```

farshidtz (changes requested):

cmd/cli/commands/get.go
```diff
@@ -43,7 +43,7 @@ func (cmd *getCommand) getValue(key string) error {
 	}
 
 	if len(value) == 0 {
-		return fmt.Errorf("no value set for key %q", key)
+		return fmt.Errorf("Key %q is not found\n\n%s", key, common.SuggestKeyNotFound(key))
```
farshidtz: Consistent with other chained errors:
```suggestion
		return fmt.Errorf("key %q is not found\n\n%s", key, common.SuggestKeyNotFound(key))
```

cmd/cli/commands/set.go
```diff
@@ -102,7 +102,7 @@ func (cmd *setCommand) setUserConfig(key, value string) error {
 	}
 	currVal, found := currValMap[key]
 	if !found && !strings.HasPrefix(key, "passthrough.") {
-		return fmt.Errorf("unknown key: %q", key)
+		return fmt.Errorf("Key %q is not found\n\n%s", key, common.SuggestKeyNotFound(key))
```
farshidtz:
```suggestion
		return fmt.Errorf("key %q is not found\n\n%s", key, common.SuggestKeyNotFound(key))
```

cmd/cli/commands/set_test.go
```diff
@@ -82,18 +83,18 @@ func TestSetValueSuccessForUserConfig(t *testing.T) {
 		t.Fatalf("setValue returned an unexpected error: %v", err)
 	}
 
-	values, err := config.Get("api.endpoint")
+	values, err := mockConfig.Get("user.api.endpoint")
```
farshidtz: This doesn't seem right. Get is expected to apply the overrides and return the final value, without using the internal `user.` prefix. The internals are subject to change and not something other packages can rely on.

pkg/storage/mock_config.go
```diff
@@ -5,45 +5,70 @@ import (
 	"strings"
 )
 
-type mockConfig struct {
+type MockConfig struct {
```
farshidtz: Should this be exported?

farshidtz (changes requested):

farshidtz (approved):

PR #339, opened 2026-04-23
Chat: use `reasoning_content` for thinking

jpm-canonical:
## Background

DeepSeek-R1 used a `<think/>` tag to show when the reasoning ended and when the main response started. The opening `<think>` tag was always missing. The `chat` was written to assume a response from DeepSeek-R1 is always started in thinking mode, and it ends when the closing tag is seen. We later removed this assumption, and required an opening `<think>` tag to trigger formatting the response as reasoning.

With Gemma-4 and Nemotron-3 we see that they do not use the think tags. They instead use a dedicated field in the response json called `reasoning_content`, with the main response content being in the `content` field.

Our webui already handled the `reasoning_content` response field correctly.

I can not find any formal documentation about this field, and the OpenAI API library we use does not implement it. Only vLLM mentions it:
https://docs.vllm.ai/en/v0.8.2/features/reasoning_outputs.html

## This PR

We drop support for the think-tags, as only DeepSeek-R1 used them (incorrectly).

We implement the `reasoning_content` field and use that to render the thinking text.

Limitations:
* Every transition from `reasoning_content` to main `content` will print a blank line. If we do not do this, there will be no space between the two. For our current models this is fine because reasoning output is always received first, and then followed by main content. If this changes, we will need to revisit this.

Tested with both Gemma-4 and Nemtron-3-nano:
Screenshot From 2026-04-23 17-23-43

### DIFF
--- a/cmd/cli/common/chat.go
+++ b/cmd/cli/common/chat.go
@@ -335,7 +335,7 @@ func (c *chatClient) processStream(stream *ssestream.Stream[openai.ChatCompletio
 				thinking = true
 				fmt.Printf("%s", color.BlueString(reasoningChunk))
 			}
-			// If the response has main content, and we were printing reasoning, add a new line
+			// If the response has main content, but we were printing reasoning, add a new line to split them
 			if thinking && contentChunk != "" {
 				thinking = false
 				fmt.Print("\n\n")
--- a/cmd/cli/common/chat.go
+++ b/cmd/cli/common/chat.go
@@ -334,15 +334,15 @@ func (c *chatClient) processStream(stream *ssestream.Stream[openai.ChatCompletio
 			if reasoningChunk != "" {
 				thinking = true
 				fmt.Printf("%s", color.BlueString(reasoningChunk))
-			} else if thinking && contentChunk != "" {
-				// Reasoning just ended, print a blank line before the main response
+			}
+			// If the response has main content, and we were printing reasoning, add a new line
+			if thinking && contentChunk != "" {
 				thinking = false
 				fmt.Print("\n\n")
 			}
 
 			// Main response
 			fmt.Printf("%s", contentChunk)
-
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/cli/common/chat.go
```diff
@@ -317,21 +316,33 @@ func (c *chatClient) processStream(stream *ssestream.Stream[openai.ChatCompletio
...
-				thinking = false
-				fmt.Printf("%s", color.BlueString(lastChunk))
+			// Gemma-4 and Nemotron-3 uses a field `reasoning_content` for thinking.
+			// The OpenAI library does not contain it, so parse it manually.
+			var rawDelta struct {
+				ReasoningContent string `json:"reasoning_content"`
+			}
+			err := json.Unmarshal([]byte(chunk.Choices[0].Delta.RawJSON()), &rawDelta)
+			if err != nil {
+				// ignore err and continue printing main content
```
farshidtz: If unmarshalling isn't known to fail (RawJSON() always giving a json), it is better to return here to ease debugging when an unexpected error happens.

cmd/cli/common/chat.go
```diff
@@ -317,21 +316,33 @@ func (c *chatClient) processStream(stream *ssestream.Stream[openai.ChatCompletio
...
 		if len(chunk.Choices) > 0 {
-			lastChunk := chunk.Choices[0].Delta.Content
 
-			if strings.Contains(lastChunk, "<think>") {
-				thinking = true
-				fmt.Printf("%s", color.BlueString(lastChunk))
-			} else if strings.Contains(lastChunk, "</think>") {
-				thinking = false
-				fmt.Printf("%s", color.BlueString(lastChunk))
+			// Gemma-4 and Nemotron-3 uses a field `reasoning_content` for thinking.
```
farshidtz:
```suggestion
			// Gemma-4 and Nemotron-3 set a `reasoning_content` field for thinking output.
```

farshidtz (changes requested):
I think deprecating <think> is reasonable. This is only used by DeepSeek R1, QwQ, and a few others that this CLI does not support.

farshidtz (approved):

PR #340, opened 2026-04-27
Remove warning printouts from library

farshidtz:
The library prints two warnings. In a few cases, the caller uses a spinner for slow operations and the printed warnings interrupt the spinner. For example:
```
$ gemma4 list-engines 
Checking engines |Warning: unable to get additional properties for pci device: Intel: getting gpu properties: looking up vram: executing clinfo: exec: "clinfo": executable file not found in $PATH
Checking engines \Warning: unable to get additional properties for pci device: NVIDIA: getting gpu properties: looking up vram: executing nvidia-smi: exit status 6: No devices were found
ENGINE      VENDOR         DESCRIPTION                                   COMPAT
nvidia-gpu  Canonical Ltd  Optimized for inference on Nvidia GPUs, usi…  yes   
cpu         Canonical Ltd  Optimized for a range of CPU micro-architec…  yes   
amd-gpu     Canonical Ltd  Use ROCm for inference on AMD GPUs.           no
```

This PR changes the library to pass warnings to the caller, to be printed at the right time after stopping spinner.

### DIFF
--- a/cmd/cli/commands/list-engines.go
+++ b/cmd/cli/commands/list-engines.go
@@ -52,14 +52,10 @@ func ListEngines(ctx *common.Context) *cobra.Command {
 }
 
 func (cmd *listEnginesCommand) run(_ *cobra.Command, _ []string) error {
-	stopProgress := common.StartProgressSpinner(common.ProgressScoring)
-	defer stopProgress()
-
-	scoredEngines, err := common.ScoreEngines(cmd.Context)
+	scoredEngines, err := common.ScoreEnginesWithSpinner(cmd.Context)
 	if err != nil {
-		return fmt.Errorf("checking engines: %v", err)
+		return fmt.Errorf("scoring engines: %v", err)
 	}
-	stopProgress()
 
 	activeEngine, err := cmd.Cache.GetActiveEngine()
 	if err != nil {
--- a/cmd/cli/commands/debug/select.go
+++ b/cmd/cli/commands/debug/select.go
@@ -62,14 +62,10 @@ func (cmd *selectCommand) run(_ *cobra.Command, args []string) error {
 		return fmt.Errorf("loading engines from directory: %s", err)
 	}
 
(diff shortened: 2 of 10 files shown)


jpm-canonical (commented):
I did a quick scan, and the change looks ok to me

cmd/cli/commands/list-engines.go
```diff
@@ -52,14 +52,10 @@ func ListEngines(ctx *common.Context) *cobra.Command {
...
 
 func (cmd *listEnginesCommand) run(_ *cobra.Command, _ []string) error {
-	stopProgress := common.StartProgressSpinner(common.ProgressScoring)
-	defer stopProgress()
-
-	scoredEngines, err := common.ScoreEngines(cmd.Context)
+	scoredEngines, err := common.ScoreEnginesWithSpinner(cmd.Context)
 	if err != nil {
-		return fmt.Errorf("checking engines: %v", err)
+		return fmt.Errorf("scoring engines: %v", err)
```
mrmara, line 57: Even though this wasn't added by this PR, since it's being modified anyway, is it worth moving it to common/suggestions?

farshidtz: I'm not sure it makes things any better. These error prefixes are identical but they don't have to be. The callers don't expect this particular prefix. It could become a reusable error but we don't need that anywhere right now. Alternatively, it could be moved to the ScoreEnginesWithSpinner function itself in favor of repeating it every time it is called, however, I prefer to leave the error context to this scope and have control over it.

mrmara (approved):

cmd/cli/commands/list-engines.go
jpm-canonical: Now that my attention got pulled here, do we want to expose the term `scoring` to the user? In the past we tried hiding it, as we didn't want to show we use `scoring` to do the selection.

PR #343, opened 2026-04-28
Make the FQDN configurable

farshidtz:
The OpenAI Base URL that is passed to the Web UI uses a hardcoded `localhost` host name. This breaks the Web UI when it is browsed on a different machine.

This PR replaces the hardcoded host name with a configurable fully qualified domain name (FQDN). To set it, the user would run `sudo <snap> set http.fqdn=<domain>`, where domain could be a .local hostname, an IP address, or a public domain name.

If `fqdn` is not set during installation, the CLI sets it to `localhost`. The same thing could be done with any generic configuration such as `host`, `verbose`, to avoid having to set it on install and refresh.

```console
$ stack-utils status
engine: intel-cpu
services:
    server: active
    server-webui: inactive
endpoints:
    kserve: http://localhost:8322/2
    openai: http://localhost:8322/v3
    openvino: http://localhost:8322/v1
    webui: http://localhost:8323/

$ stack-utils get
http.fqdn: localhost
http.host: 127.0.0.1
http.port: 8322
verbose: 3
webui.http.host: 127.0.0.1
webui.http.port: 8323

$ sudo stack-utils set http.fqdn=test
Restart stack-utils to apply the changes? [Y/n] Y

$ stack-utils status
engine: intel-cpu
services:
    server: active
    server-webui: inactive
endpoints:
    kserve: http://test:8322/2
    openai: http://test:8322/v3
    openvino: http://test:8322/v1
    webui: http://localhost:8323/

$ curl -s http://localhost:8323/config | jq
{
  "openAIBaseURL": "http://test:8322/v3",
  "capabilities": [
    "text",
    "vision"
  ],
  "instanceName": "stack-utils",
  "engineName": "intel-cpu"
}
```

### DIFF
--- a/cmd/cli/common/endpoints.go
+++ b/cmd/cli/common/endpoints.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
+
+	"github.com/canonical/inference-snaps-cli/pkg/storage"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


farshidtz:
Closing:
- FQDN is not well known, localhost (the default) is not a valid FQDN
- FQDN may be confused with the `host` field (bind address)
- Web UI can construct the OpenAI base URL using `window.location.hostname`

The /config endpoint needs to split the URL segments in `openAIBaseURL` to pass port and basepath separately.

PR #346, opened 2026-04-29
Support setting multiple configurations at once

farshidtz:
Add support for setting multiple key-values with the `set` command.


mrmara:
Tested, working as expected!

mrmara (approved):

PR #350, opened 2026-04-30
Fix refresh when there is no active engine

mrmara:
This PR addresses an issue where the post-refresh hook fails under specific conditions, blocking the refresh process.

Currently, if a snap lacks auto-connection permissions or is installed locally, the use-engine command invoked during the post-refresh hook fails because it cannot locate an active engine.

Modified the `use-engine` command invoked with  `--fix` flag to handle cases where cli cannot find an active engine implementing a silent failure mechanism for this specific command, ensuring that the hook continues execution even if there is no active engine

### DIFF
--- a/cmd/cli/commands/use-engine_test.go
+++ b/cmd/cli/commands/use-engine_test.go
@@ -65,6 +65,6 @@ func TestFixActiveEngine_noActiveEngine(t *testing.T) {
 
 	err := cmd.fixActiveEngine()
 	if !errors.Is(err, common.ErrNoActiveEngine) {
-		t.Errorf("expected ErrNoActiveEngine, got %v", err)
+		t.Errorf("expected no active engine error, got %v", err)
 	}
 }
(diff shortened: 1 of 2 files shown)


cmd/cli/commands/use-engine_test.go
```diff
@@ -49,3 +52,19 @@ func ExampleUseEngine_restartWhenEngineChanged() {
...
+		Context: &common.Context{
+			EnginesDir: "../../../test_data/engines",
+			Cache:      cache,
+			Snap:       snap.Mock(),
+		},
+	}
+
+	err := cmd.fixActiveEngine()
+	if !errors.Is(err, common.ErrNoActiveEngine) {
+		t.Errorf("expected ErrNoActiveEngine, got %v", err)
```
farshidtz: There is a high likelihood of code references in logs, error messages, and comments becoming out of sync with the code itself.
```suggestion
	if !errors.Is(err, common.ErrNoActiveEngine) {
		t.Errorf("expected no active engine error, got %v", err)
```

farshidtz (approved):

PR #351, opened 2026-04-30
Add snapshot tests for show-engine

imatrisciano:
Added snapshot tests for show-engine, targeting both `json` and `yaml` as output formats

### DIFF
--- a/cmd/cli/commands/show-engine_test.go
+++ b/cmd/cli/commands/show-engine_test.go
@@ -1,9 +1,12 @@
 package commands
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/canonical/inference-snaps-cli/pkg/engines"
+	"github.com/canonical/inference-snaps-cli/pkg/hardware_info"
+	"github.com/canonical/inference-snaps-cli/pkg/selector"
 )
 
 func TestInfoLong(t *testing.T) {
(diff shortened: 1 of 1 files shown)


cmd/cli/commands/show-engine_test.go
```diff
@@ -37,3 +40,133 @@ func TestInfoShort(t *testing.T) {
...
+		panic(fmt.Sprintf("failed to load engine manifest: %v", err))
+	}
+	info, err := hardware_info.GetFromRawData("dummy-machine", true, "../../../test_data")
+	if err != nil {
+		panic(fmt.Sprintf("failed to get hardware info: %v", err))
+	}
+	scoredEngines, err := selector.ScoreEngines(info, []engines.Manifest{*engineManifest})
+	if err != nil {
+		panic(fmt.Sprintf("failed to score engines: %v", err))
+	}
```
farshidtz: Consider moving these to a function that takes the engine as input and returns a scoredEngine.

farshidtz (changes requested):
@imatrisciano thanks for the added coverage. Could you also add positive tests (no top level `compatibility-issues`)?

mrmara:
Thanks, Ivano. IMO, it would be great to add coverage for the `switch-default` case in `printEngineManifest` in this PR, even if it’s not a snapshot test.
image

mrmara (approved):

farshidtz (approved):

PR #352, opened 2026-05-04
Introduce list engine snapshot tests

imatrisciano:
This PR adds snapshot tests to catch any unexpected change to the `list-engines` command output. Both the json and the table output formats are tested.

Please note that the Table output format test does not use a standard _Example_ test as the test framework doesn't accept trailing whitespaces for the expected output, and the test would fail. A standard unit test was written instead.


cmd/cli/commands/list-engines.go
```diff
@@ -97,7 +99,9 @@ func (cmd *listEnginesCommand) printEnginesJson(enginesList outputEngines) error
 	return nil
 }
 
-func (cmd *listEnginesCommand) printEnginesTable(enginesList outputEngines) error {
+var noEnginesFoundError = errors.New("No engines found.")
+
+func (cmd *listEnginesCommand) getEnginesTable(enginesList outputEngines) (*string, error) {
```
farshidtz: Pointer to a string is not idiomatic in Go. The function can return an empty string on error.
```suggestion
func (cmd *listEnginesCommand) getEnginesTable(enginesList outputEngines) (string, error) {
```

cmd/cli/commands/list-engines.go
```diff
@@ -207,15 +210,31 @@ func (cmd *listEnginesCommand) printEnginesTable(enginesList outputEngines) erro
...
+
+func (cmd *listEnginesCommand) printEnginesTable(enginesList outputEngines) error {
+	tableOutput, err := cmd.getEnginesTable(enginesList)
+	if err != nil {
+		if errors.Is(err, noEnginesFoundError) {
+			fmt.Fprintln(os.Stderr, "No engines found.")
+			return nil
+		}
+		return fmt.Errorf("generating table: %v", err)
+	}
```
farshidtz, line 234: Perhaps a bit more readable without the nested if statement:
```suggestion
	if errors.Is(err, noEnginesFoundError) {
		fmt.Fprintln(os.Stderr, "No engines found.")
		return nil
	} else if err != nil {
		return fmt.Errorf("generating table: %v", err)
	}
```

But why not just check the length of enginesList before calling this function?

cmd/cli/commands/list-engines_test.go
```diff
@@ -54,13 +55,1228 @@ func TestList(t *testing.T) {
...
+	}
+
+	err = cmd.printEnginesJson(*enginesList)
+	if err != nil {
+		panic(fmt.Sprintf("Error printing engines json: %v", err))
+	}
+
+	// Output:
+	// {
+	//   "active-engine": "example-memory",
```
farshidtz, line 141: This (>1000 lines) is going to be a pain to maintain. How about reducing it to a couple items?

```suggestion
	err = cmd.printEnginesJson(*enginesList[0:1])
	if err != nil {
		panic(fmt.Sprintf("Error printing engines json: %v", err))
	}

	// Output:
	// {
	//   "active-engine": "example-memory",
```

farshidtz (changes requested):
Thanks for extending the tests. I've commented on a few things inline.

cmd/cli/commands/list-engines_test.go
imatrisciano: That's true, I've also noticed that the output was huge, but I didn't think about forcing it onto a subset of engines.

I've introduced this change, but instead of picking the first engines in the list, I've added a filtering method that accepts a list of engine names as a whitelist. This way the test will not fail if new engines are added, or the order of existing engines changes

cmd/cli/commands/list-engines.go
imatrisciano: I have simplified the nested if statements as you've suggested.

Regarding the `No engines found.` situation, I agree that it could be simplified by checking the length of `engineList` before attempting to call the print function, but the original implementation only checked for the engine count after the table was built and I didn't want to change that. Maybe the table generation used to do some extra filtering while listing the engines, and this is just a leftover? If so I can quickly refactor it to clean it up

farshidtz: Looking deeper inside the function, I see slice indexing (`tableRows[1:]`) that could panic, so it is wise to check the size before reading that. 

However, I think the new printEnginesJson function doesn't need to pass an empty slice to getEnginesTable and rely on a new error type to capture that the slice was empty. It feels over-engineered to me, but I'd leave it up to you to decide.

mrmara:
tested, working as expected!

mrmara (approved):

farshidtz (approved):

PR #354, opened 2026-05-04
Add snapshot tests for autoselect engine

mrmara:
Snapshot tests validate a method's output by comparing it against a stored reference standard. This reference must be an exact match with the generated output to pass. To ensure consistency across both local and remote environments, a new test data directory has been established containing three engines designed to support the specifications of most machines.


farshidtz (changes requested):
> To ensure consistency across both local and remote environments, a new test data directory has been established containing three engines designed to support the specifications of most machines.

Relying on the host machine makes these pretty fragile. What if this is run on a machine with Ampere Altra? Instead of limiting the references, I suggest making autoSelectEngine more testable by taking the scored engines as input.

cmd/cli/commands/use-engine_test.go
```diff
@@ -53,6 +56,57 @@ func ExampleUseEngine_restartWhenEngineChanged() {
...
+	cpuDevelManifest, err := engines.LoadManifest(cmd.Context.EnginesDir, "cpu-devel")
+	if err != nil {
+		panic(err)
+	}
+	cpuManifest, err := engines.LoadManifest(cmd.Context.EnginesDir, "cpu")
+	if err != nil {
+		panic(err)
+	}
+
+	allEngines := []engines.Manifest{*notCompatibleManifest, *cpuDevelManifest, *cpuManifest}
```
farshidtz: A bit more extensible?
```suggestion
var allEngines []engines.Manifest
for _, name := range []string{"not-compatible-engine", "cpu-devel", "cpu"} {
	e, err := engines.LoadManifest(cmd.Context.EnginesDir, name)
	if err != nil {
		panic(err)
	}
	allEngines = append(allEngines, *e)
}
```

cmd/cli/commands/use-engine.go
```diff
@@ -100,6 +100,11 @@ func (cmd *useEngineCommand) autoSelectEngine() error {
 		return fmt.Errorf("scoring engines: %v", err)
 	}
 
+	return cmd.autoSelectEngineWithScoredEngines(scoredEngines)
+}
+
+func (cmd *useEngineCommand) autoSelectEngineWithScoredEngines(scoredEngines []engines.ScoredManifest) error {
```
farshidtz:
```suggestion
func (cmd *useEngineCommand) autoSelectScoredEngine(scoredEngines []engines.ScoredManifest) error {
```

farshidtz (approved):

PR #355, opened 2026-05-05
Add bash completion for configuration values

mrmara:
commands: `set`,`get` and `unset` will autocomplete known configuration variables.

Special case:
If setting multiple variable in the same line, already completed vars will be excluded from further completion


cmd/cli/commands/config_completion.go
```diff
@@ -0,0 +1,35 @@
...
+		}
+
+		if strings.HasPrefix(candidate, toComplete) {
+			completions = append(completions, candidate)
+		}
+	}
+
+	sort.Strings(completions)
+	return completions, cobra.ShellCompDirectiveNoFileComp
+}
```
imatrisciano: Can `excludedKeys` an array of strings instead of a map? I feel this would make the function more readable

cmd/cli/commands/set.go
```diff
@@ -56,6 +56,26 @@ func (cmd *setCommand) run(_ *cobra.Command, args []string) error {
...
+		if idx := strings.IndexByte(arg, '='); idx >= 0 {
+			key = arg[:idx]
+		}
+		if key != "" {
+			excludedKeys[key] = struct{}{}
+		}
+	}
+
+	completions, directive := completeConfigKeys(cmd.Config, toComplete, true, excludedKeys)
+	return completions, directive | cobra.ShellCompDirectiveNoSpace
```
imatrisciano: The `cobra.ShellCompDirectiveNoSpace` directive is not working for me.

When I type `stack-utils set http.h` and press [TAB] it gets completed to `stack-utils set http.host= `, so a space is added at the end. I need to press backspace so the syntax can be correct

imatrisciano (changes requested):
The `cobra.ShellCompDirectiveNoSpace` directive is not working for me.

When I type `stack-utils set http.h` and press [TAB] it gets completed to `stack-utils set http.host= `, so a space is added at the end. I need to press backspace so the syntax can be correct

cmd/cli/commands/config_completion.go
mrmara: Yes it is possible, but making it an array of strings would require to iterate for each configuration over the array each time, multiple times. IMO, It does not make the code much more readable.

cmd/cli/commands/set.go
mrmara: I am experiencing the very same behavior. I checked and cobra is doing the same for any other tab completion. Seems that `cobra.ShellCompDirectiveNoSpace` has no effect

imatrisciano: This issue exists both in bash and in zsh

farshidtz:
I can confirm that the extra space is only added when using the bash completion override in the snap.

Without it, it works as expected:
```console
$ git diff
diff --git a/snap/snapcraft.yaml b/snap/snapcraft.yaml
index a51f63e..a336df5 100644
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -160,7 +160,7 @@ components:
 apps:
   stack-utils:
     command: bin/modelctl
-    completer: bin/completion.bash
+    # completer: bin/completion.bash
     plugs:
       - hardware-observe
       - home


# build and install

$ stack-utils set we^C                   # no completion
$ source <(stack-utils completion bash)  # source the completion script
$ stack-utils set webui.http.host=^C     # completion without an extra space
```

cmd/cli/common/config_completion.go
farshidtz, line 1: Please move this to common. Files under `/commands` strictly map to the CLI commands.

pkg/storage/mock_config.go
```diff
@@ -62,9 +61,21 @@ func (c *mockConfig) Get(key string) (map[string]any, error) {
...
-	maps.Copy(allValues, c.values)
-	return allValues, nil
+	result := make(map[string]any)
+
+	for _, confType := range []configType{PackageConfig, EngineConfig, UserConfig} {
+		prefix := string(confType) + "."
+		for fullKey, value := range c.values {
+			if !strings.HasPrefix(fullKey, prefix) {
+				continue
+			}
```
farshidtz, line 71: In what case a prefix would not be present?

farshidtz (changes requested):
For the record, tab completion has been removed for the `set` command because of an incompatibility in the completion script, which is beyond the scope of this PR.

pkg/storage/mock_config.go
mrmara: This check is needed to respect hierarchy when getting same key set at multiple level, hence following (in order of precedence) User->Engine->Package

farshidtz (approved):

PR #356, opened 2026-05-05
Report model name in status

jpm-canonical:
Source the model name from the env vars exported in the component.yaml files. If it is set, report the model name in the status yaml.

```
$ stack-utils status
engine: intel-cpu
services:
    server: active
    server-webui: active
endpoints:
    kserve: http://localhost:8322/2
    openai: http://localhost:8322/v3
    openvino: http://localhost:8322/v1
    webui: http://localhost:8323/
model:
    name: gemma-3-4b-it-ov-int4-fq
```

Or if no model name is set in the component yamls:
```
$ stack-utils status
engine: cpu-avx1
services:
    server: active
    server-webui: inactive
```

### DIFF
--- a/cmd/cli/common/model_status.go
+++ b/cmd/cli/common/model_status.go
@@ -1,0 +1,35 @@
+package common
+
+import (
+	"fmt"
+	"strings"
+)
+
+func ModelStatus(ctx *Context) (map[string]string, error) {
+	settings, err := EngineComponentSettings(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("loading engine environment: %v", err)
+	}
+	return modelStatus(settings)
+}
+
+func modelStatus(settingsCollection []ComponentSettings) (map[string]string, error) {
+	status := make(map[string]string)
+	for _, settings := range settingsCollection {
+		for i := range settings.Environment {
+			// Split into key/value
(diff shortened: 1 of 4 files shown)


cmd/cli/common/model_status.go
farshidtz, line 1: The `common` package is primarily used for implementation that is common to multiple commands but does not fit into pkg, e.g. because they print to standard streams or are closely coupled to the CLI logic. It looks to me that this can be added directly in status.go, unless you foresee shared usage between commands.

farshidtz (approved):
Thanks, I tested with gemma4 and it works as expected:
```
$ gemma4 status
engine: nvidia-gpu
services:
    server: active
    server-webui: active
endpoints:
    openai: http://localhost:8336/v1
    webui: http://localhost:8337/
model:
    name: gemma4-e4b-q4-k-m
```

cmd/cli/common/model_status.go
jpm-canonical: I think it is nicer having it in the common package, as then it follows the same pattern as `ServerEndpoints`. Same pattern, i.e. easier to maintain. And indeed, `ModelStatus` might be called elsewhere in the future, just like `ServerEndpoints` is called in chat.

PR #361, opened 2026-05-20
Shorten environment passthrough config prefix

mrmara:
The prefix to set environment variables configurations has been simplified from `passthrough.environment.` to `env.`.


farshidtz (approved):
Looks good to me.
```console
$ sudo stack-utils set env.aaa=1
Restart stack-utils to apply the changes? [Y/n] 
$ stack-utils run env | grep AAA
AAA=1
$ sudo stack-utils set env.aaa=2
Restart stack-utils to apply the changes? [Y/n] 
$ stack-utils run env | grep AAA
AAA=2
```

PR #364, opened 2026-05-27
Replace engine grade with "Experimental"

mrmara:
Engine `grade [stable|devel]` has been changed to a boolean filed: `Experimental`.

If not set, false is assumed


cmd/cli/commands/list-engines.go
```diff
@@ -110,7 +110,10 @@ func (cmd *listEnginesCommand) getEnginesTable(enginesList outputEngines) (strin
 	sort.Slice(enginesList.Engines, func(i, j int) bool {
 		// Stable engines with equal score should be listed first
 		if enginesList.Engines[i].Score == enginesList.Engines[j].Score {
-			return enginesList.Engines[i].Grade == "stable"
+			if enginesList.Engines[i].Experimental == nil {
+				return false
+			}
+			return *enginesList.Engines[i].Experimental == false
```
farshidtz: I see this and similar logic (`engine.Experimental != nil && *engine.Experimental`) in a few places. Consider adding a function `engine.IsExperimental()` that checks and returns a bool.

README.md
```diff
@@ -91,7 +91,7 @@ $ inference-snaps-cli show-machine --format=json | inference-snaps-cli debug sel
 ✅ cpu-avx1 - compatible, score = 14
 ✅ cpu-avx2 - compatible, score = 17
 ❌ cpu-avx512 - not compatible: required cpu device not found
-🟠 cpu-devel - score = 12, grade = devel
+🟠 cpu-devel - score = 12, experimental=true
```
farshidtz: Is this up to date?
```suggestion
🟠 cpu-devel - score = 12, experimental
```

cmd/cli/commands/list-engines_test.go
```diff
@@ -144,7 +144,7 @@ func Example_printEnginesJson() {
 	//       "name": "amd-gpu",
 	//       "description": "AMD specific engine targeting only one microarchitecture.",
 	//       "vendor": "Canonical Ltd",
-	//       "grade": "stable",
+	//       "experimental": false,
```
farshidtz: I'm not in favor of the inconsistency exposed to the user, where experimental is shown when explicitly set to false but not when left unset, which can coexist in a single snap. 

Ideally, this should always be hidden (look into deleting false bool pointers in LoadManifest(s)).

farshidtz (changes requested):
Nice, this is much clearer than "devel" which conflicts with the grade of the snap itself.

I've commented on a few minor things.

farshidtz (approved):

PR #365, opened 2026-05-28
Add summary in engine manifest

mrmara:
Add to the engine manifest a compulsory `summary` field with a limit of 56 characters. 56 is the maximum amount of characters that can be shown in `list-engines` table provided the minimum width for any other column.

`description` field changed to optional and omitted if empty


farshidtz:
@mrmara could you rebase this on the main branch?

README.md
```diff
@@ -91,7 +91,7 @@ $ inference-snaps-cli show-machine --format=json | inference-snaps-cli debug sel
 ✅ cpu-avx1 - compatible, score = 14
 ✅ cpu-avx2 - compatible, score = 17
 ❌ cpu-avx512 - not compatible: required cpu device not found
-🟠 cpu-devel - score = 12, grade = devel
+🟠 cpu-devel - score = 12, experimental=true
```
jpm-canonical: Maybe we should rename this engine too
```suggestion
🟠 cpu-exptl - score = 12, experimental=true
```

pkg/engines/validate.go
```diff
@@ -77,8 +77,12 @@ func (manifest Manifest) validate(expectedEngineName string) error {
...
 	}
 
-	if manifest.Description == "" {
-		return fmt.Errorf("required field is not set: description")
+	if manifest.Summary == "" {
+		return fmt.Errorf("required field is not set: summary")
+	} else if len(manifest.Summary) > 56 {
+		// 56 limit has been chosen as it is the maximum length that can be displayed in the list-engines table
+		// without truncation considering the minimum width of the other columns.
+		return fmt.Errorf("summary field should be at most 56 characters: %d", len(manifest.Summary))
```
farshidtz: Ideally, move 56 to a constant along with the comment. Don't repeat the value in the comment itself because it is an added maintenance burden to keep them in sync.

To make sure this value is changed according to the list-engine's table changes, consider adding an artificial check in the table generation code. 


```suggestion
	} else if len(manifest.Summary) > MaxSummaryLength {
		return fmt.Errorf("summary field should be at most %s characters, got: %d", MaxSummaryLength, len(manifest.Summary))
```

pkg/engines/validate_test.go
```diff
@@ -39,12 +40,21 @@ func TestManifestFiles(t *testing.T) {
...
+					err = Validate(manifestPath)
+					if err != nil {
+						t.Fatalf("%s: %v", engine, err)
+					}
+				})
+			} else {
+				t.Run(engine, func(t *testing.T) {
+					err = Validate(manifestPath)
+					if err == nil {
+						t.Fatalf("%s should be invalid because too summary exceeds length", engine)
```
farshidtz:
```suggestion
						t.Fatalf("%s should be invalid because the summary exceeds length", engine)
```

test_data/engines/amd-gpu/engine.yaml
```diff
@@ -1,5 +1,7 @@
 name: amd-gpu
-description: AMD specific engine targeting only one microarchitecture.
+description: AMD specific engine targeting 
+  only one microarchitecture.
```
farshidtz:
```suggestion
description: | 
  AMD specific engine targeting one microarchitecture:
    - the arch
```

And make sure the output (show-engine) maintains line breaks.

farshidtz (changes requested):
Thanks. The addition of summary in place of description is to avoid ellipses in the table, while supporting rich engine descriptions with line-break aware formatting

pkg/engines/validate.go
mrmara: Done. I added a warning in case the engine description column max length is bigger than the allowed maximum

cmd/cli/commands/list-engines.go
```diff
@@ -148,6 +149,9 @@ func (cmd *listEnginesCommand) getEnginesTable(enginesList outputEngines) (strin
 	engineVendorMaxLen += 2
 	// Description column fills the remaining space
 	engineDescriptionMaxLen := tableMaxWidth - (engineNameMaxLen + engineVendorMaxLen)
+	if engineDescriptionMaxLen > engines.SummaryMaxLength {
```
farshidtz: Please refactor the variable names: description->summary. And log/error messages.

cmd/cli/commands/list-engines.go
```diff
@@ -146,10 +147,13 @@ func (cmd *listEnginesCommand) getEnginesTable(enginesList outputEngines) (strin
 	// Increase column widths to account for paddings
 	engineNameMaxLen += 1
 	engineVendorMaxLen += 2
-	// Description column fills the remaining space
-	engineDescriptionMaxLen := tableMaxWidth - (engineNameMaxLen + engineVendorMaxLen)
+	// Summary column fills the remaining space
+	engineSummaryMaxLen := tableMaxWidth - (engineNameMaxLen + engineVendorMaxLen)
+	if engineSummaryMaxLen > engines.SummaryMaxLength {
+		fmt.Printf("Warning: summary column max length %d is greater than summary max length %d, setting it to summary max length\n", engineSummaryMaxLen, engines.SummaryMaxLength)
+	}
```
jpm-canonical: The only goal of this if statement is to print a warning. Does this warning add any improvements to the user experience? I expect the validation to detect too long summaries, and by the time we get here, if the summary is still too long, it will just be truncated with an ellipse. Is a warning really needed?

jpm-canonical (approved):
Looks good to me. I'm just wondering if the warning that is printed is really needed.

cmd/cli/commands/list-engines.go
mrmara: For the same reasons you mentioned, I thought an error would be too much. The check has to be performed, and the warning seemed harmless. But I agree that it's not adding value to the UX

farshidtz: The goal was to inform the developer that a maximum length is defined somewhere else, calculated based on the logic of the code written here. The max may require modification based on the changes to this table. We're in this strange situation because the table generation is in the main package, and the constant cannot be defined here, where it belongs, and referenced during the validation. 

An ugly solution could be:
```suggestion
    # Note: summary length has a hard limit
	_ := engines.SummaryMaxLength
```

A cleaner one would be to just comment on it and hope that the developer reads it (they should):
`// Summary column fills the remaining space, up to [engines.SummaryMaxLength]`

See https://tip.golang.org/doc/comment#links

farshidtz:
@mrmara Please resolve the conflicts.

dev
farshidtz, line 1: This change is unexpected.

farshidtz (approved):

PR #369, opened 2026-06-03
Support setting a fallback when auto selecting or fixing an engine

mrmara:
This PR refactors duplicated and brittle logic from the install and post-refresh hooks across inference snaps. Previously, failed hardware observability checks in these hooks could break installations by failing to set an active engine.

To fix this, the use-engine command has been updated to:
- Handle hardware observability checks directly
- Accept a new `--fallback=<engine-name>`  flag to set a default engine if hardware is not observable.


cmd/cli/commands/use-engine.go
```diff
@@ -22,10 +22,21 @@ type useEngineCommand struct {
...
 
+
+var isHardwareObserveConnected = func() (bool, error) {
+	return snapctl.IsConnected("hardware-observe").Run()
+}
+
+var canAccessHardwareInfo = func() bool {
+	_, err := os.ReadDir("/sys/bus/pci/devices")
+	return err == nil
+}
```
farshidtz: I understand that function pointers are set to a global variable to allow overriding in tests. Instead, please move these functions to the pkg/snap package and make them mockable.

farshidtz (changes requested):
Thanks. Nice simplification of the hooks.

cmd/cli/commands/use-engine.go
```diff
@@ -95,6 +111,17 @@ func (cmd *useEngineCommand) run(_ *cobra.Command, args []string) error {
...
+	if cmd.fallback != "" {
+		connected, err := isHardwareObserveConnected()
+		if err != nil {
+			return fmt.Errorf("checking hardware-observe connection: %v", err)
+		}
+		if !connected && !canAccessHardwareInfo() {
+			fmt.Printf("Hardware information is unavailable; falling back to engine %q.\n", cmd.fallback)
+			return cmd.switchEngine(cmd.fallback)
+		}
+	}
```
farshidtz: Consider abstracting away this complexity:
```suggestion
	if snap.HardwareObservable() && cmd.fallback != "" {
		fmt.Printf("Hardware information is unavailable; falling back to engine %q.\n", cmd.fallback)
		return cmd.switchEngine(cmd.fallback)
	}
```

HardwareObservable():
  - hardware-observe connection
  - file access (dev mode, classic snaps)

cmd/cli/commands/use-engine.go
```diff
@@ -111,15 +101,9 @@ func (cmd *useEngineCommand) run(_ *cobra.Command, args []string) error {
...
-	if cmd.fallback != "" {
-		connected, err := isHardwareObserveConnected()
-		if err != nil {
-			return fmt.Errorf("checking hardware-observe connection: %v", err)
-		}
-		if !connected && !canAccessHardwareInfo() {
-			fmt.Printf("Hardware information is unavailable; falling back to engine %q.\n", cmd.fallback)
-			return cmd.switchEngine(cmd.fallback)
-		}
+	if !cmd.Context.Snap.HardwareObservable() && cmd.fallback != "" {
```
farshidtz: Snap is "inherited" from Context (embedded into useEngineCommand).
```suggestion
	if !cmd.Snap.HardwareObservable() && cmd.fallback != "" {
```

cmd/cli/commands/use-engine.go
```diff
@@ -70,6 +72,10 @@ func (cmd *useEngineCommand) run(_ *cobra.Command, args []string) error {
 		return common.ErrPermissionDenied
 	}
 
+	if cmd.fallback != "" && !cmd.auto {
+		return fmt.Errorf("--fallback requires --auto")
```
farshidtz:
```suggestion
		return fmt.Errorf("--fallback must be used together with --auto")
```

pkg/snap/snap.go
```diff
@@ -29,3 +32,15 @@ func (*snap) Restart(service ...string) error {
...
+func (*snap) HardwareObservable() bool {
+	connected, err := snapctl.IsConnected("hardware-observe").Run()
+	if err != nil {
+		return false
+	}
+	_, err = os.ReadDir("/sys/bus/pci/devices")
+	if err != nil {
+		return false
+	}
+	return (connected && (err == nil))
```
farshidtz: The code already returns early if err != nil.

```suggestion
	return connected
```

pkg/snap/snap.go
```diff
@@ -29,3 +32,15 @@ func (*snap) Restart(service ...string) error {
...
+
+func (*snap) HardwareObservable() bool {
+	connected, err := snapctl.IsConnected("hardware-observe").Run()
+	if err != nil {
+		return false
+	}
+	_, err = os.ReadDir("/sys/bus/pci/devices")
+	if err != nil {
+		return false
+	}
```
farshidtz, line 51: I suggest returning the errors to the caller to capture unexpected situations, e.g. failing to check hardware observer or not finding any pci/devices (different from a permission error).

pkg/snap/snap.go
```diff
@@ -29,3 +32,15 @@ func (*snap) Restart(service ...string) error {
 func (*snap) InstanceName() string {
 	return env.SnapInstanceName()
 }
+
+func (*snap) HardwareObservable() bool {
+	connected, err := snapctl.IsConnected("hardware-observe").Run()
+	if err != nil {
+		return false
+	}
```
farshidtz, line 43: Return early. There is no need to check pci devices explicitly if the interface is already connected, unless we want to go ahead and check every path that we may need to be failsafe.

```suggestion
	}
	if connected {
		return true
	}
```

farshidtz (changes requested):

farshidtz (approved):

PR #376, opened 2026-06-09
Implement `list-models` and `show-model`

jpm-canonical:
This can be tested using https://github.com/canonical/gemma4-snap/pull/39

Added:

- [x] `list-models` prints a table list list-engines
- [x] `list-models --format=json`
- [x] `show-model <model-name|model-id>` prints the manifest in yaml
- [x] `show-model <model-name|model-id> --format=json`

Tested:

- [x] Install from branch, auto selection: intel-gpu
- [x] `list-models` only show one model compatible with intel-gpu engine
- [x] `show-model` no params
- [x] `show-model` auto-complete
- [x] Switch engine to: cpu
- [x] `list-models` shows three compatible models
- [x] `show-model` auto-completes three models

Issues/questions:

* ~~`show-model` prints the entire model manifest as is. This also includes the `layout`. Should we hide the `layout` from this output?~~
* `status` reports the model name, as should be provided when using the OpenAI API. This name is different from the one reported by `list-models` and `show-model`


cmd/cli/commands/list-models_test.go
```diff
@@ -0,0 +1,112 @@
...
+		t.Fatalf("Error preparing test data: %v", err)
+	}
+
+	err = cmd.printModelsTable(*modelsList)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func Example_printModelsJson() {
```
farshidtz, line 99: Please add tests for the list-models table output.

cmd/cli/commands/list-models.go
```diff
@@ -51,11 +64,156 @@ func (cmd *listModelsCommand) run(_ *cobra.Command, _ []string) error {
 		return fmt.Errorf("%s: %w", common.LoadingEngineManifest, err)
 	}
 
+	var modelsList outputModels
 	for _, model := range engineManifest.Model.Options {
-		// TODO in IENG-2398: handle json and table formats, including model metadata
-		// For now just print a list of model IDs
-		fmt.Println(model)
+		modelManifest, err := models.LoadManifest(cmd.ModelsDir, model)
```
farshidtz, line 69: Disk size here and in show-model can benefit from reformatting to the largest number group.

I'm getting:
```
NAME  CAPABILITIES  DISK SIZE                                                   
e2b   text          2963M                                                       
e4b*  text          5089M                                                       
26b   text          16163M 
```

cmd/cli/commands/show-model.go
```diff
@@ -0,0 +1,144 @@
...
+
+	var manifest *models.Manifest
+	for i := range manifests {
+		if manifests[i].Name == modelName || manifests[i].ID == modelName {
+			manifest = &manifests[i]
+			break
+		}
+	}
+	if manifest == nil {
+		return fmt.Errorf(`model "%s" does not exist`, modelName)
```
farshidtz:
```suggestion
		return fmt.Errorf("model %q does not exist", modelName)
```

farshidtz (changes requested):
Thanks. I've tested this with the suggested gemma4 build. 

> show-model prints the entire model manifest as is. This also includes the layout. Should we hide the layout from this output?

Indeed, attributes that belong to the packaging (layout, environment) should not be displayed in show/list-models. Introduce a decoupled data model (ModelDetails, similar to the existing EngineDetails) that inherits and overrides?

> status reports the model name, as should be provided when using the OpenAI API. This name is different from the one reported by list-models and show-model

Ideally, the model name in the API should be taken from the same source, but I'm afraid this isn't always feasible. I believe once we deprecate `MODEL_NAME` with something like `modelctl show-model | jq .name`, this will end up in sync for servers that support setting the model name at runtime. 

Thinking about this further, maybe the name here should have been "alias" as in our early design. This also solves the problem of these names not being unique, unlike other names in the snap ecosystem.

```
$ gemma4 show-model
id: e4b-it-int4-ov
name: e4b
description: E4B model for OpenVINO
model-card-url: https://huggingface.co/OpenVINO/gemma-4-E4B-it-int4-ov
quantization: INT4
capabilities:
    - text
disk-size: 6161M
components:
    - model-e4b-it-int4-ov-1-of-2
    - model-e4b-it-int4-ov-2-of-2
environment:
    - MODEL_NAME=gemma4-e4b-it-int4-ov
    ...
layout:
    ...
```

jpm-canonical:
> Thinking about this further, maybe the name here should have been "alias" as in our early design. This also solves the problem of these names not being unique, unlike other names in the snap ecosystem.

I agree. Let's do that in a separate PR, as it would require updating the test data and the consuming snap's model manifests.

farshidtz (approved):

PR #380, opened 2026-06-17
Rename `cli` to `modelctl` and restructure release artifacts

imatrisciano:
This PR renames `cli` into `modelctl` and includes the snap-compatible bash completion script in release assets to improve re-usability of the same.

The new release asset's file structure looks like this:

- bin/
   - modelctl
   - snap-completer.bash
- LICENSE


.github/workflows/release-artifacts.yaml
```diff
@@ -49,11 +49,11 @@ jobs:
 
       - name: Generate artifact
         run: |
-          tar -Jcvf ${{ env.ARTIFACT_NAME }} ./cli ./LICENSE
+          tar -Jcvf ${{ env.ARTIFACT_NAME }} ./cli ./LICENSE ./scripts/snap-completion.bash
```
farshidtz: The "scripts" directory grouping is superfluous. Why not keep it alongside the cli?

completer/snap.bash
farshidtz, line 1: Having this script at the top level of a Go module is misleading. The command that this script runs doesn't belong to the module, but is rather something that gets created for the package.

farshidtz: Maybe the cli itself should be renamed to modelctl.

.github/workflows/release-artifacts.yaml
imatrisciano: Addressed by placing both under `bin/`

completer/snap.bash
imatrisciano: I've moved the completion script under a dedicated `completer` directory, and renamed the `cli` in `modelctl`

farshidtz:
```suggestion
# Bash completion script for modelctl packaged in a snap
#
# To use, in snapcraft.yaml:
# 1. Set apps.<app-name>.completer to the name of this script for the app.
# 2. Add a symlink from <app> to modelctl binary.
#
# For example, snapcraft.yaml:
# name: mymodel
# <...>
# parts:
#    cli:
#      source: <...>
#      plugin: dump
#      override-build: |
#        # For tab completion
#        ln --symbolic ./modelctl bin/mymodel
#        craftctl default
#      apps:
#        mymodel: # command named after the snap
#          command: bin/modelctl
#          completer: bin/snap-completer.bash

```

Please ensure formatting.

farshidtz: > Please ensure formatting.

Formatting please.

imatrisciano: Sorry about that, I didn't notice `apps` wasn't in the correct indentation level and focused on tab width consistency. Should be corrected now

farshidtz (approved):

PR #385, opened 2026-06-19
Installing snap with components overrides engine selection

jpm-canonical:
This PR can be tested using `gemma4` from `latest/edge/pr-69`. A summary of tests that were done on it can be seen in [the PR description](https://github.com/canonical/gemma4-snap/pull/69).

The following has not been implemented exactly as specified. The current behaviour is that all seeded runtimes and all seeded models translate to a list of engines. These engines are sorted by score, and the highest scoring engine from this list is selected.

> If more than one model or more than one runtime is installed with the snap, ignore all. This is to ensure that the system doesn’t change the selection mechanism when the user’s intention is to simply have multiple things installed to switch over later (e.g. UC image with several components for offline use).

### Breaking changes

* Runtime manifests require an `id` field, with the value matching the runtime directory name.

### DIFF
--- a/cmd/modelctl/commands/use-engine.go
+++ b/cmd/modelctl/commands/use-engine.go
@@ -339,7 +339,18 @@ func (cmd *useEngineCommand) migrateConfig() error {
 	return nil
 }
 
+// engineNames extracts the Name field from a slice of engine manifests using the provided accessor.
+func engineNames[T any](items []T, getName func(T) string) []string {
+	names := make([]string, len(items))
+	for i, item := range items {
+		names[i] = getName(item)
+	}
+	return names
+}
+
 func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []engines.ScoredManifest) (bool, error) {
+	fmt.Println("Checking preinstalled components to influence engine and model selection")
+
 	allEngines, err := engines.LoadManifests(cmd.EnginesDir)
 	if err != nil {
 		return false, err
@@ -358,6 +369,8 @@ func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []e
 		return false, err
 	}
 
+	fmt.Printf("Installed components: %v\n", installedComponents)
+
 	var seededRuntimes []string
 	var seededModels []string
 
@@ -374,6 +387,7 @@ func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []e
 			seededRuntimes = append(seededRuntimes, runtime.ID)
 		}
 	}
+	fmt.Printf("Seeded runtimes: %v\n", seededRuntimes)
 
 	for _, model := range allModels {
 		// check if all the model.Components are included in installedComponents. If it is, add the model ID to seededModels
@@ -388,6 +402,7 @@ func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []e
 			seededModels = append(seededModels, model.ID)
 		}
 	}
+	fmt.Printf("Seeded models: %v\n", seededModels)
 
 	// iterate all engines and make a list of ones for which the runtime is in seededRuntimes, or at least one of its model options is in seededModels.
 	// Engines where both conditions are true are preferred over engines where only one condition is true.
@@ -408,6 +423,8 @@ func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []e
 			partiallySeededEngines = append(partiallySeededEngines, engine)
 		}
 	}
+	fmt.Printf("Partially seeded engines: %v\n", engineNames(partiallySeededEngines, func(e engines.Manifest) string { return e.Name }))
+	fmt.Printf("Fully seeded engines: %v\n", engineNames(fullySeededEngines, func(e engines.Manifest) string { return e.Name }))
 
 	// filterCompatible returns the subset of the given engines that have a score >0 in scoredEngines.
 	filterCompatible := func(candidates []engines.Manifest) []engines.ScoredManifest {
@@ -428,9 +445,11 @@ func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []e
 	if len(compatibleSeededEngines) == 0 {
 		compatibleSeededEngines = filterCompatible(partiallySeededEngines)
 	}
+	fmt.Printf("Compatible seeded engines: %v\n", engineNames(compatibleSeededEngines, func(e engines.ScoredManifest) string { return e.Name }))
 
 	// Seeded components do not target any compatible engine
 	if len(compatibleSeededEngines) == 0 {
+		fmt.Printf("No compatible seeded engines found; falling back to standard auto selection.\n")
 		return false, nil
 	}
 
@@ -439,6 +458,7 @@ func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []e
 	if err != nil {
 		return false, fmt.Errorf("finding top engine: %v", err)
 	}
+	fmt.Printf("Top engine: %v\n", topEngine.Name)
 
 	// at this point we need to also know which one of the seeded models is the one from model.options of the topEngine
 	// If there are multiple matches, prefer the default model
@@ -451,10 +471,14 @@ func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []e
 			}
 		}
 	}
-
-	err = cmd.Cache.SetActiveModel(seededModelForEngine)
-	if err != nil {
-		return false, fmt.Errorf("setting active model: %v", err)
+	fmt.Printf("Seeded model for engine: %v\n", seededModelForEngine)
+	// If one of the components defined a supported model, set it as the active model.
+	// Otherwise leave the active model unset so that the default can be set by switchEngine()
+	if seededModelForEngine != "" {
+		err = cmd.Cache.SetActiveModel(seededModelForEngine)
+		if err != nil {
+			return false, fmt.Errorf("setting active model: %v", err)
+		}
 	}
 
 	err = cmd.switchEngine(topEngine.Name)
--- a/cmd/modelctl/commands/use-engine.go
+++ b/cmd/modelctl/commands/use-engine.go
@@ -375,16 +375,12 @@ func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []e
 	var seededModels []string
 
 	for _, runtime := range allRuntimes {
-		// check if all the runtime.Components are included in installedComponents. If it is, add the runtime name to seededRuntimes
-		allInstalled := true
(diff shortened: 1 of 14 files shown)


pkg/runtimes/types.go
```diff
@@ -3,6 +3,7 @@ package runtimes
 import "github.com/canonical/inference-snaps-cli/v2/pkg/types"
 
 type Manifest struct {
+	ID          string                  `json:"id" yaml:"id"`
 	Servers     map[string]Server       `json:"servers" yaml:"servers"`
 	Environment []string                `json:"environment" yaml:"environment"`
 	Layout      map[string]types.Layout `yaml:"layout"`
```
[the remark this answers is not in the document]
jpm-canonical: I believe the `cli debug validate-engines` command should do this validation. The validation should validate all engine manifests, and recurse down into the listed runtime and models, and validate their manifests too. This is however a bigger change that is not related to this PR, so I suggest we do it separately.

test_data/runtimes/llama-cpp-rocm/runtime.yaml
```diff
@@ -12,4 +14,4 @@ environment:
   - LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$SNAP_COMPONENTS/llama-cpp-cuda/usr/lib/$ARCH_TRIPLET
 
 components:
-  - llama-cpp-cuda
+  - runtime-llama-cpp-rocm
```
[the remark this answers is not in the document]
jpm-canonical: @copilot check again

test_data/runtimes/llama-cpp-cuda/runtime.yaml
```diff
@@ -12,4 +14,4 @@ environment:
   - LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$SNAP_COMPONENTS/llama-cpp-cuda/usr/lib/$ARCH_TRIPLET
 
 components:
-  - llama-cpp-cuda
+  - runtime-llama-cpp-cuda
```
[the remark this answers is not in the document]
jpm-canonical: @copilot check again

test_data/runtimes/llama-cpp-cpu/runtime.yaml
```diff
@@ -1,3 +1,5 @@
+id: llama-cpp-cpu
```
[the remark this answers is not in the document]
jpm-canonical: @copilot check again

cmd/modelctl/commands/use-engine.go
```diff
@@ -45,6 +47,7 @@ func UseEngine(ctx *common.Context) *cobra.Command {
 	cobraCmd.Flags().StringVar(&cmd.fallback, "fallback", "", "fallback engine to use when hardware information is unavailable (requires --auto or --fix)")
 	cobraCmd.Flags().BoolVar(&cmd.assumeYes, "assume-yes", false, "assume yes for all prompts")
 	cobraCmd.Flags().BoolVar(&cmd.noRestart, "no-restart", false, "do not restart the snap after changing engine")
+	cobraCmd.Flags().BoolVar(&cmd.considerComponents, "components", false, "consider pre-installed components")
```
farshidtz:
```suggestion
	cobraCmd.Flags().BoolVar(&cmd.considerComponents, "components", false, "consider pre-installed components (requires --auto)")
```

Also, add a check to error out when used without --auto.

cmd/modelctl/commands/use-engine.go
```diff
@@ -141,6 +144,20 @@ func (cmd *useEngineCommand) autoSelectScoredEngine(scoredEngines []engines.Scor
...
+		switched, err := switchToPreinstalledEngineAndModel(cmd, scoredEngines)
+		if err != nil {
+			return err
+		}
+		if switched {
+			// switched, so return success
+			return nil
+		}
+		// otherwise continue with standard auto selection
+	}
```
farshidtz: If "switched" needs a comment to clarify that it means success:
```suggestion
		success, err := switchToPreinstalledEngineAndModel(cmd, scoredEngines)
		if err != nil {
			return err
		}
		if success {
			return nil
		}
		// otherwise continue with standard auto selection
	}
```

cmd/modelctl/commands/use-engine.go
```diff
@@ -321,3 +338,145 @@ func (cmd *useEngineCommand) migrateConfig() error {
...
+// engineNames extracts the Name field from a slice of engine manifests using the provided accessor.
+func engineNames[T any](items []T, getName func(T) string) []string {
+	names := make([]string, len(items))
+	for i, item := range items {
+		names[i] = getName(item)
+	}
+	return names
+}
+
+func switchToPreinstalledEngineAndModel(cmd *useEngineCommand, scoredEngines []engines.ScoredManifest) (bool, error) {
```
farshidtz: Regarding the function name: an engine is not pre-installed. Consider `selectEngineForSeededComponents` which also matches the implementation.

cmd/modelctl/commands/use-engine.go
```diff
@@ -321,3 +338,145 @@ func (cmd *useEngineCommand) migrateConfig() error {
...
+	if err != nil {
+		return false, err
+	}
+
+	fmt.Printf("Installed components: %v\n", installedComponents)
+
+	var seededRuntimes []string
+	var seededModels []string
+
+	// A runtime or model is considered seeded if any one required component for the respective runtime or model is currently installed.
```
farshidtz: A little wordy. 
```suggestion
	// A runtime or model is considered seeded if any of its components are currently installed
```

cmd/modelctl/commands/use-engine.go
```diff
@@ -321,3 +338,145 @@ func (cmd *useEngineCommand) migrateConfig() error {
...
+			if slices.Contains(installedComponents, component) {
+				seededModels = append(seededModels, model.ID)
+				break
+			}
+		}
+	}
+	fmt.Printf("Seeded models: %v\n", seededModels)
+
+	// iterate all engines and make a list of ones for which the runtime is in seededRuntimes, or at least one of its model options is in seededModels.
+	// Engines where both conditions are true are preferred over engines where only one condition is true.
```
farshidtz: Please shorten. To stay valuable and easy to maintain, comments should avoid repeating the logic and variable names.

cmd/modelctl/commands/use-engine.go
```diff
@@ -321,3 +338,145 @@ func (cmd *useEngineCommand) migrateConfig() error {
...
+			}
+		}
+		if hasSeededRuntime && hasSeededModel {
+			fullySeededEngines = append(fullySeededEngines, engine)
+		} else if hasSeededRuntime || hasSeededModel {
+			partiallySeededEngines = append(partiallySeededEngines, engine)
+		}
+	}
+	fmt.Printf("Partially seeded engines: %v\n", engineNames(partiallySeededEngines, func(e engines.Manifest) string { return e.Name }))
+	fmt.Printf("Fully seeded engines: %v\n", engineNames(fullySeededEngines, func(e engines.Manifest) string { return e.Name }))
```
farshidtz: Line breaks can improve code readability.

cmd/modelctl/commands/use-engine.go
```diff
@@ -321,3 +338,145 @@ func (cmd *useEngineCommand) migrateConfig() error {
...
+
+	// If multiple seeded engines, find the top one
+	topEngine, err := selector.TopEngine(compatibleSeededEngines)
+	if err != nil {
+		return false, fmt.Errorf("finding top engine: %v", err)
+	}
+	fmt.Printf("Top engine: %v\n", topEngine.Name)
+
+	// at this point we need to also know which one of the seeded models is the one from model.options of the topEngine
+	// If there are multiple matches, prefer the default model
```
farshidtz: Not sure I understand the first line.
```suggestion
	// If there are multiple matches, prefer the default model
```

cmd/modelctl/commands/use-engine.go
```diff
@@ -321,3 +338,145 @@ func (cmd *useEngineCommand) migrateConfig() error {
...
+	// Otherwise leave the active model unset so that the default can be set by switchEngine()
+	if seededModelForEngine != "" {
+		fmt.Printf("Seeded model for engine: %v\n", seededModelForEngine)
+		err = cmd.Cache.SetActiveModel(seededModelForEngine)
+		if err != nil {
+			return false, fmt.Errorf("setting active model: %v", err)
+		}
+	}
+
+	err = cmd.switchEngine(topEngine.Name)
```
farshidtz: This is assuming switchEngine respects the currently selected active model. This can change independantly. Perhaps make switchEngine aware of this, with an explicit input.

cmd/modelctl/commands/use-engine_test.go
```diff
@@ -215,3 +262,209 @@ func TestFixActiveEngine_fallbackWhenActiveEngineMissing(t *testing.T) {
...
+// CPU + Intel GPU (where intel-gpu normally wins on score alone):
+//
+//	installed component               expected engine   expected model
+//	──────────────────────────────────────────────────────────────────
+//	model-26b-a4b-q4-k-m-gguf         cpu               26b-q4-k-m-gguf
+//	mmproj-26b-bf16-gguf               cpu               26b-q4-k-m-gguf
+//	model-4b-it-int4-fq-ov             intel-gpu         4b-it-int4-fq-ov
+//	runtime-llama-cpp-cpu              cpu               26b-q4-k-m-gguf (default)
+//	model-30b-a3b-q4-k-m-gguf-3-of-6  cpu               30b-a3b-q4-k-m-gguf
+//	runtime-llama-cpp-cuda (incompat.) –                 – (falls back, returns false)
```
farshidtz: Redundant comments, hard to keep up to date.

pkg/runtimes/types.go
```diff
@@ -3,6 +3,7 @@ package runtimes
 import "github.com/canonical/inference-snaps-cli/v2/pkg/types"
 
 type Manifest struct {
+	ID          string                  `json:"id" yaml:"id"`
```
farshidtz: I understand this is consistent with the model manifest and how both IDs end up in engine's manifest. But I'm concerned with added use of IDs as identifiers, where snap, components, engines utilize names. 

Elsewhere in the code we consider runtime's name as their identifier: https://github.com/canonical/inference-snaps-cli/blob/0c8b8f5376e3d6eb1eb6d8859db102fad94cc203/cmd/cli/common/component.go#L149

How about we stick with name and separately reconsider use of IDs for models. Perhaps use "name" for the identifier and alias/label for the friendly name?

farshidtz (approved):
Looks good overall, combined with the testing done with gemma4: https://github.com/canonical/gemma4-snap/pull/69

I've commented a few minor things.

cmd/modelctl/commands/use-engine.go
jpm-canonical: Yes, that would be better. But it touches a wider range of code. Let me rather do it separate from this PR.

PR #386, opened 2026-06-23
Fix model lookup for the active engine

jpm-canonical:
`show-model` now only takes models into account that are compatible with the active engine. The lookup from model name/id is done by a function shared between show-model and use-model.

```
$ stack-utils list-models
NAME     CAPABILITIES               DISK SIZE                                   
26b*     text                       6G                                          
30b-a3b  text, vision, audio, tool  6G                                          
$
$ sudo stack-utils use-model 30b-a3b 
Model changed to "30b-a3b".
Restart stack-utils to apply the changes? [Y/n] 
$
$ stack-utils list-models
NAME      CAPABILITIES               DISK SIZE                                  
26b       text                       6G                                         
30b-a3b*  text, vision, audio, tool  6G                                         
$
$ sudo stack-utils use-model 26b
Model changed to "26b".
Restart stack-utils to apply the changes? [Y/n] 
$
$ stack-utils list-models
NAME     CAPABILITIES               DISK SIZE                                   
26b*     text                       6G                                          
30b-a3b  text, vision, audio, tool  6G                                          
$
$ sudo stack-utils use-model 4b
Error: model "4b" is not compatible with active engine
$
$ sudo stack-utils use-model 4a
Error: model "4a" does not exist
```


cmd/modelctl/commands/show-model.go
```diff
@@ -99,19 +105,60 @@ func (cmd *showModelCommand) showCurrentModel() error {
...
 		}
 	}
-	if manifest == nil {
+
+	var manifest *models.Manifest
+	switch {
+	case idMatch != nil:
+		manifest = idMatch
+	case nameMatches > 1:
+		return fmt.Errorf("multiple compatible models are named %q; please use the model ID instead", modelName)
```
farshidtz: We should make sure this never happens, otherwise, list-models will contain indistinguishable entries. What would be the best way to do that?

cmd/modelctl/commands/show-model.go
```diff
@@ -18,6 +20,10 @@ type showModelCommand struct {
 
 	// flags
 	format string
+
+	// out is the writer used by printModelManifest. When nil it falls back to os.Stdout,
+	// which keeps the Example tests working and lets unit tests inject a bytes.Buffer.
+	out io.Writer
```
farshidtz: This is different from how we've been testing printed outputs. We typically use Go's testing [Examples](https://pkg.go.dev/testing#hdr-Examples) when we really need to test printouts. For function-level unit testing, we should really be testing small functional units that return and expect the caller to print.

cmd/modelctl/commands/show-model_test.go
```diff
@@ -72,3 +99,301 @@ func Example_showModelCommand_printModelManifestJson() {
...
+	}
+	output := buf.String()
+
+	// model-b was matched by ID; its unique name should appear, not model-a's.
+	if !strings.Contains(output, "model-b-name") {
+		t.Errorf("expected output to contain ID-matched model name %q, got:\n%s", "model-b-name", output)
+	}
+	if strings.Contains(output, "model-a") {
+		t.Errorf("expected output NOT to contain name-matched model %q, got:\n%s", "model-a", output)
+	}
```
farshidtz: These tests can potentially match some other string and pass.

Same comment applies to 2 other tests.

cmd/modelctl/commands/show-model_test.go
```diff
@@ -72,3 +99,301 @@ func Example_showModelCommand_printModelManifestJson() {
...
+		format:  "yaml",
+	}
+
+	err := cmd.showModel("shared-name")
+	if err == nil {
+		t.Fatal("expected an error for ambiguous model name, got nil")
+	}
+	if !strings.Contains(err.Error(), "model ID") {
+		t.Errorf("expected error to mention %q, got: %v", "model ID", err)
+	}
```
farshidtz: Can this look for something more specific that indicates an error?

farshidtz (approved):
LGTM. Thanks for the fix.

Tested with gemma4.

PR #387, opened 2026-06-24
Allow model override in `switchEngine`

jpm-canonical:
This PR addresses the concern from https://github.com/canonical/inference-snaps-cli/pull/385#discussion_r3460600996

Switch engine should be explicitly told which model to switch to in some situations. We should not depend on the active model being set as having this as a side effect.

### DIFF
--- a/cmd/modelctl/commands/use-engine.go
+++ b/cmd/modelctl/commands/use-engine.go
@@ -177,7 +177,7 @@ func (cmd *useEngineCommand) autoSelectScoredEngine(scoredEngines []engines.Scor
 // switchEngine changes the engine that is used by the snap
 // By default the previous model will be used if it is compatible.
 // If it is not compatible, the engine's default model will be selected.
-// And optional modelName can be provided to override which model is switched to.
+// An optional modelName can be provided to override the model used for the switch.
 func (cmd *useEngineCommand) switchEngine(engineName string, modelName string) error {
 
 	newEngineManifest, err := engines.LoadManifest(cmd.EnginesDir, engineName)
--- a/cmd/modelctl/commands/use-engine.go
+++ b/cmd/modelctl/commands/use-engine.go
@@ -101,7 +101,7 @@ func (cmd *useEngineCommand) run(_ *cobra.Command, args []string) error {
 		return err
 	} else {
 		if len(args) == 1 {
-			return cmd.switchEngine(args[0])
+			return cmd.switchEngine(args[0], "")
 		} else {
 			return fmt.Errorf("engine name not specified")
 		}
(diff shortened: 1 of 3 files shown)


cmd/modelctl/commands/use-engine.go
```diff
@@ -188,24 +191,28 @@ func (cmd *useEngineCommand) switchEngine(engineName string) error {
...
-	newModelName := activeModelName
-	if !slices.Contains(newEngineManifest.Model.Options, activeModelName) {
-		newModelName = newEngineManifest.Model.Default
+	// If a model ID is provided, use that instead
+	if modelID != "" {
+		if !slices.Contains(newEngineManifest.Model.Options, modelID) {
+			return fmt.Errorf("model %q is not supported by engine %q", modelID, engineName)
+		}
+		newModelID = modelID
 	}
```
farshidtz: Not concerned with efficiency, but I find this a bit hard to follow. This is how I read the code:
1. new model is active model
2. no, new model is active model only if it is compatible, otherwise default is the new model
3. no, the given model is the new model

How about:

```go
    var newModelID string
	if modelID == "" {
		if slices.Contains(newEngineManifest.Model.Options, activeModelID) {
			newModelID := activeModelID
		} else {
			newModelID := newEngineManifest.Model.Default
		}
	} else { 
		if !slices.Contains(newEngineManifest.Model.Options, modelID) {
			return fmt.Errorf("model %q is not supported by engine %q", modelID, engineName)
		}
		newModelID = modelID
	}
```
or a slightly more convoluted one:
```go
    newModelID := newEngineManifest.Model.Default
	if modelID == "" {
		if slices.Contains(newEngineManifest.Model.Options, activeModelID) {
			newModelID := activeModelID
		}
	} else { 
		if !slices.Contains(newEngineManifest.Model.Options, modelID) {
			return fmt.Errorf("model %q is not supported by engine %q", modelID, engineName)
		}
		newModelID = modelID
	}
```

cmd/modelctl/commands/use-engine.go
```diff
@@ -175,7 +175,10 @@ func (cmd *useEngineCommand) autoSelectScoredEngine(scoredEngines []engines.Scor
 }
 
 // switchEngine changes the engine that is used by the snap
-func (cmd *useEngineCommand) switchEngine(engineName string) error {
+// By default the previous model will be used if it is compatible.
+// If it is not compatible, the engine's default model will be selected.
+// An optional modelID can be provided to override the model used for the switch.
+func (cmd *useEngineCommand) switchEngine(engineName string, modelID string) error {
```
farshidtz: Most calls to this function do not pass a model ID. How keeping the original signature for a new wrapper function that doesn't take modelID, and renaming this one to `switchEngineWithModelID`?

farshidtz (changes requested):

cmd/modelctl/commands/use-engine.go
jpm-canonical: Great idea. By introducing the wrapper function, we can solve the other comment about readability too. The auto-selection of a model is now in the wrapper, making the main `switchEngineAndModel()` much cleaner.

farshidtz: But it is now instead making the same calls to query active engine and read manifest twice in a row. Calls to snapd typically take a negligible time, around 50-100ms to respond, but these all pile up as there are usually several others to account for.

farshidtz: Let's keep this as-is since the function signatures are quite neat, and we separately try to make sure that calls to read "cache" aren't slow.

farshidtz (approved):

PR #388, opened 2026-06-24
Reject dots in environment variable names

mrmara:
By accepting environment variables names with dots the CLI produce and export environment varibales not POSIX compliant.

Now dots and trailng dots are no more accepted.

```
$ sudo stack-utils set env.a.b=c
Error: invalid key "env.a.b": dots are not allowed in environment variable names
$ sudo stack-utils set env.ab=c
Restart stack-utils to apply the changes? [Y/n] n
```


farshidtz (approved):
LGTM. Nice added testing coverage.

PR #389, opened 2026-06-29
Add `export-status` subcommand

jpm-canonical:
The server script calls `modelctl export-status`, which updates the config shared via the content sharing interfaces.

The legacy `open-webui` interface, sharing a `openai.json` files is still in place, using the same directory as the `status` interface.


cmd/modelctl/commands/export-config.go
```diff
@@ -0,0 +1,98 @@
...
+
+	statusJson, err := json.Marshal(sharedStatusStr)
+	if err != nil {
+		return fmt.Errorf("json: %v", err)
+	}
+
+	shareDir := os.Getenv("STATUS_SHARE_DIR")
+	if shareDir == "" {
+		shareDir = filepath.Join(os.Getenv("SNAP_COMMON"), "share")
+	}
```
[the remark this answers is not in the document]
jpm-canonical: This is ok as this application will only be used from inside a snap.

cmd/modelctl/commands/status.go
```diff
@@ -39,7 +38,7 @@ func Status(ctx *common.Context) *cobra.Command {
 		"yaml",
 		fmt.Sprintf("output format (%s)", strings.Join(supportedFormats, ", ")),
 	)
-	cobraCmd.Flags().BoolVar(&cmd.waitForComponents, "wait-for-components", false, "wait for engine components to be installed before reporting status")
+	cobraCmd.Flags().MarkDeprecated("wait-for-components", "\"status\" never waits for components.")
```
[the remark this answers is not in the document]
jpm-canonical: Reverting this change as it should be done in a separate PR

snap/local/server.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash -eu
 
+modelctl export-config
+
```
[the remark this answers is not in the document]
jpm-canonical: I think it is ok for the script to exit if this command fails, otherwise issues are hidden.

cmd/modelctl/commands/export-config.go
```diff
@@ -0,0 +1,111 @@
...
+		Services:  statusStr.Services,
+		Endpoints: statusStr.Endpoints,
+		Model:     statusStr.Model,
+	}
+
+	// Default to $SNAP_COMMON/share unless it is overridden by $STATUS_SHARE_DIR
+	shareDir := os.Getenv("STATUS_SHARE_DIR")
+	if shareDir == "" {
+		shareDir = filepath.Join(os.Getenv("SNAP_COMMON"), "share")
+	}
```
farshidtz: Please make this a CLI argument for better visibility and help.

Need to check the spec to verify if arg 0 can be made optional. The serve-webui has it as mandatory, and I think that could be made optional too with a reasonable default.

cmd/modelctl/common/status.go
```diff
@@ -0,0 +1,47 @@
...
+	Services  map[string]string `json:"services" yaml:"services"`
+	Endpoints map[string]string `json:"endpoints,omitempty" yaml:"endpoints,omitempty"`
+	Model     map[string]string `json:"model,omitempty" yaml:"model,omitempty"`
+}
+
+// serviceStatusesFn is the function used to retrieve service statuses.
+// It can be overridden in tests.
+var serviceStatusesFn = ServiceStatuses
+
+func StatusStruct(ctx *Context) (*Status, error) {
```
farshidtz: The original name wasn't ideal. Let's improve that since this is now a public function.

cmd/modelctl/commands/export-config.go
```diff
@@ -0,0 +1,111 @@
...
+		Hidden:            true,
+		Args:              cobra.NoArgs,
+		ValidArgsFunction: cobra.NoFileCompletions,
+		RunE:              cmd.run,
+	}
+
+	return cobraCmd
+}
+
+type sharedStatus struct {
```
farshidtz: It might be less confusing (and better scoped considering that this is a package with another command called status) if this matched the command terminology. 

Instead of renaming this type, I suggest changing the command and having it named after this (`share-status`?) since the content interface also use "status" for the slot.

cmd/modelctl/commands/export-config.go
```diff
@@ -0,0 +1,111 @@
...
+	Engine    string            `json:"engine" yaml:"engine"`
+	Services  map[string]string `json:"services" yaml:"services"`
+	Endpoints map[string]string `json:"endpoints,omitempty" yaml:"endpoints,omitempty"`
+	Model     map[string]string `json:"model,omitempty" yaml:"model,omitempty"`
+}
```
farshidtz: I think we can exclude engine and services until a use case demands it. Especially services is something we can't share accurately. 

Does the model need to be a map? Not sure why status uses a map either.

cmd/modelctl/common/status_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/canonical/inference-snaps-cli/v2/pkg/storage"
+)
+
+func TestStatusStruct(t *testing.T) {
+	// Stub out ServiceStatuses so the test never calls snapctl.
+	orig := serviceStatusesFn
+	defer func() { serviceStatusesFn = orig }()
+	serviceStatusesFn = func() (map[string]string, error) {
+		return map[string]string{"llama-server": "active"}, nil
+	}
```
farshidtz: Please extend and use https://github.com/canonical/inference-snaps-cli/blob/main/pkg/snap/mock_snap.go for tests.

cmd/modelctl/common/status_test.go
```diff
@@ -0,0 +1,94 @@
...
+	modelsDir := t.TempDir()
+
+	writeEngineYAML(t, enginesDir, "my-engine", `name: my-engine
+runtime: my-runtime
+`)
+	writeRuntimeYAML(t, runtimesDir, "my-runtime", `servers:
+  openai:
+    protocol: http
+    base-path: /v1
+`)
```
farshidtz, line 23: These functions are borrowed from endpoint tests. I see this is also done in model_test.go. Move to common_test.go?

cmd/modelctl/commands/export-config.go
```diff
@@ -0,0 +1,111 @@
...
+
+// sharedOpenai is to be deprecated in favor of sharedStatus.
+// The Open WebUI snap first needs to be updated to consume sharedStatus
```
farshidtz: Marking it as deprecated helps with the maintenance: https://go.dev/wiki/Deprecated

farshidtz (changes requested):
Thanks.
Nice to have the deprecated parts ported to Go, which simplifies the snap packages and their maintenance. I've commented on a few things.

cmd/modelctl/commands/export-config.go
jpm-canonical: Model is a map because it can contain multiple entries: name, quantization, etc.

cmd/modelctl/commands/export-config.go
jpm-canonical: I did one pass of renaming. Not sure I covered everything you had in mind. Please check again.

cmd/modelctl/commands/export-config.go
jpm-canonical: Done. `serve-webui` command not touched.

cmd/modelctl/commands/export-config.go
farshidtz: Right. I was wondering why it is a map and not a struct.

cmd/modelctl/commands/export-status.go
```diff
@@ -0,0 +1,110 @@
...
+type exportStatusCommand struct {
+	*common.Context
+}
+
+func ExportStatus(ctx *common.Context) *cobra.Command {
+	var cmd exportStatusCommand
+	cmd.Context = ctx
+
+	cobraCmd := &cobra.Command{
+		Use:               "export-status",
```
farshidtz:
```suggestion
		Use:               "export-status [<dir>]",
```

cmd/modelctl/commands/export-status.go
```diff
@@ -0,0 +1,110 @@
...
+		shareDir = args[0]
+	} else {
+		shareDir = filepath.Join(os.Getenv("SNAP_COMMON"), "share")
+	}
+
+	return writeShareFiles(sharedStatusStr, shareDir)
+}
+
+// writeShareFiles writes status.json (and optionally openai.json) to shareDir.
+func writeShareFiles(status *exportedStatus, shareDir string) error {
```
farshidtz: This should be a method of the command struct for encapsulation within this package.

cmd/modelctl/commands/export-config.go
farshidtz: It is okay. Could do another pass once the deprecated content sharing is removed.

farshidtz (approved):

PR #394, opened 2026-07-08
Add lint-package debug command for engines, runtimes and models

mrmara:
`lint-package` command checks and validate manifests file for engines, runtimes and models

Example usage:
`go run ./cmd/modelctl debug lint-package <snap-directory>`

Test:

Succesful
```
go run ./cmd/modelctl debug lint-package ../qwen3.5
✅ ../qwen3.5/engines/cpu/engine.yaml
✅ ../qwen3.5/engines/nvidia-gpu/engine.yaml
✅ ../qwen3.5/models/4b-q4-k-xl-gguf/model.yaml
✅ ../qwen3.5/models/9b-q4-k-m-gguf/model.yaml
✅ ../qwen3.5/runtimes/llamacpp/runtime.yaml
✅ ../qwen3.5/runtimes/llamacpp-cuda/runtime.yaml
```

Fail since `models`, `runtimes` nor `engines` directories are found:
```
go run ./cmd/modelctl debug lint-package .
❌ engines: no manifests found
❌ models: no manifests found
❌ runtimes: no manifests found
Error: some manifests are invalid
exit status 1
```

Fail since `models` directory is not found:
```
go run ./cmd/modelctl debug lint-package ../qwen3.5-another/
✅ ../qwen3.5-another/engines/cpu/engine.yaml
✅ ../qwen3.5-another/engines/nvidia-gpu/engine.yaml
❌ ../qwen3.5-another/models: no manifests found
✅ ../qwen3.5-another/runtimes/llamacpp/runtime.yaml
✅ ../qwen3.5-another/runtimes/llamacpp-cuda/runtime.yaml
Error: some manifests are invalid
exit status 1
```

Fail because of validation error on a manifest:
```
go run ./cmd/modelctl debug lint-package ../qwen3.5-another/
✅ ../qwen3.5-another/engines/cpu/engine.yaml
✅ ../qwen3.5-another/engines/nvidia-gpu/engine.yaml
❌ ../qwen3.5-another/models/qwen-3-5-4b-q4-k-xm/model.yaml: model directory name should match name in manifest: qwen-3-5-4b-q4-k-xm != Qwen 3.5 4B
✅ ../qwen3.5-another/models/qwen-3-5-9b-q4-k-m/model.yaml
✅ ../qwen3.5-another/runtimes/llamacpp/runtime.yaml
✅ ../qwen3.5-another/runtimes/llamacpp-cuda/runtime.yaml
Error: some manifests are invalid
exit status 1
```

[...]

PR #395, opened 2026-07-08
Get model capabilities from metadata

mrmara:
Model capabilities are now retrieved from metadata inside `model.yaml` and passed to webui, dropping the need for hardcoded values.

Supported capabilities are validated at build time for each model against a known list ([PR](https://github.com/canonical/inference-snaps-cli/pull/394)).

Frontend will ignore and not show any unsupported capability ([PR](https://github.com/canonical/inference-snaps-webui/pull/129))

### DIFF
--- a/pkg/webui/config.go
+++ b/pkg/webui/config.go
@@ -12,17 +12,6 @@ type Config struct {
 	EngineName    string   `json:"engineName"`
 }
 
-const (
-	capabilityText     string = "text"
-	capabilityVision   string = "vision"
-	capabilityTools    string = "tools"
-	capabilityThinking string = "thinking"
-)
-
-func SupportedCapabilities() []string {
-	return []string{capabilityText, capabilityVision, capabilityTools, capabilityThinking}
-}
-
 func (c Config) Validate() error {
 
 	// Validate OpenAI base URL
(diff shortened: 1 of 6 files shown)


pkg/webui/config_test.go
```diff
@@ -4,23 +4,6 @@ import (
 	"testing"
 )
```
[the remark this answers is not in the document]
mrmara: SupportedCapabilites() is only returning a slice of strings. No relevant tests can be performed IMO

pkg/webui/config.go
```diff
@@ -30,12 +30,8 @@ func (c Config) Validate() error {
...
 	}
 
-	// Validate capabilities
-	for _, cap := range c.Capabilities {
-		if !slices.Contains(SupportedCapabilities(), cap) {
-			return fmt.Errorf("unsupported capability: %q", cap)
-		}
-	}
+	// Capabilities are forwarded as-is; the frontend ignores any it does not
+	// recognize, so unknown capabilities are not rejected here.
```
farshidtz: In this case, why are the capabilities still defined in this package?

mrmara: moved to common package in `model.go`

farshidtz (approved):

PR #405, opened 2026-07-15
Move several snap-related operations to snap package

mrmara:
Moved every `snap` and `snapctl` related operations inside snap package

### DIFF
--- a/pkg/snap/mock_snap.go
+++ b/pkg/snap/mock_snap.go
@@ -49,6 +49,10 @@ func (c *mockSnap) InstallComponent(name string) error {
 	return nil
 }
 
+func (c *mockSnap) RemoveComponents(name ...string) error {
+	return nil
+}
+
 func (c *mockSnap) ServiceStatuses() (map[string]string, error) {
 	if c.serviceStatuses != nil {
 		return c.serviceStatuses, nil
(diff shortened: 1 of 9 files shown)


pkg/snap/snap.go
```diff
@@ -36,9 +39,39 @@ func (*snap) Restart(service ...string) error {
...
+
+// Dir returns the snap mount directory ($SNAP), or empty when not running as a snap.
+func (*snap) Dir() string {
+	return Dir()
+}
+
+// Version returns the snap version.
+func (*snap) Version() string {
+	return Version()
+}
```
farshidtz, line 53: I'm wondering why don't these methods directly call the env methods and instead rely on separate wrapper functions?

pkg/snap/mock_snap.go
```diff
@@ -38,6 +38,14 @@ func (c *mockSnap) InstanceName() string {
 	return "mock-snap"
 }
 
+func (c *mockSnap) Dir() string {
+	return ""
```
farshidtz: Tests and other code may end up using the mock directory and access/modify root-level files. See main.go (Dir()+/models, etc). 

```suggestion
	return "/tmp/mock-snap"
```

farshidtz (changes requested):

farshidtz (approved):
LGTM.
The added snap mocking should ease testing. Some mock functions can produce mocked output such as printouts so they can be relied on from consumer unit tests.

PR #412, opened 2026-07-22
feat!: Refactor URLs in status and extend protocol support

farshidtz:
Change status data model from:
```
$ gemma4 status
engine: nvidia-gpu
services:
    server: active
    server-webui: active
endpoints:
    openai: http://127.0.0.1:8336/v1
    webui: http://127.0.0.1:8337/
model:
    name: gemma4-e4b-q4-k-m
```

To:
```
$ gemma4 status
engine: nvidia-gpu
services:
    server: active
    server-webui: active
entrypoints:
    openai:
        url: http://127.0.0.1:8336/v1
    webui:
        url: http://127.0.0.1:8337/
```

And add support for "http+ws", "ws",  "ws+unix" protocols in runtime manifests. Example status:
```
engine: cpu
services:
    llama-server: active
entrypoints:
    logger:
        url: http://localhost:8081/
    openai:
        url: http://0.0.0.0:8080/v1
    whisperlive:
        unix-socket: /run/whisper.sock (ws://unix/realtime)
```

Example status in JSON:
```json
{
   "engine": "cpu",
   "services": {
    "llama-server": "active"
  },
  "entrypoints": {
    "logger": {
      "url": "http://localhost:8081/"
    },
    "openai": {
      "url": "http://0.0.0.0:8080/v1"
    },
    "whisperlive": {
      "unix-socket": "/run/whisper.sock",
      "unix-socket-url": "ws://unix/realtime"
    }
  },
  "model": {
    "name": "gemma-4-4b-it-int4-fq-ov"
  }
}
```

Add support for config namespacing in runtime manifests to allow the definition of servers listening on separate ports.
```yaml
servers:
  openai:
    protocol: http
    base-path: /v1
  kserve:
    protocol: http
    base-path: /v2
  whisperlive:
    protocol: ws+unix
    base-path: /realtime
  logger:  
    protocol: http
    base-path: /
    namespace: logger # new field
```
The namespace is used to retrieve the configurations for a server entry to construct the corresponding entrypoint. In the above example, the `openai` and `kserve` servers URLs will be constructed by using the `http.port` and `http.host` configurations (no namespace). For logger, the configuration will be taken from `logger.http.port` and `logger.http.host` instead.

**BREAKING CHANGE: This is a breaking change for the clients that rely on the `endpoints.<key>` field in status. They should instead query `entrypoints.<key>.url`.**


cmd/modelctl/common/endpoints.go
```diff
@@ -34,75 +56,142 @@ func ServerEndpoints(ctx *Context) (map[string]string, error) {
...
 			}
-			endpoints[serverName] = httpUrl
 		default:
-			return nil, fmt.Errorf("unsupported protocol %q for server %q in component %q",
+			return nil, fmt.Errorf("unsupported protocol %q for server %q in runtime %q",
 				serverSettings.Protocol, serverName, activeEngineManifest.Runtime)
 		}
+
+		entrypoints[serverName] = *entrypoint
 	}
```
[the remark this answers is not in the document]
farshidtz: I can't skip it if it is nil. It needs to be either an error with sensible details for debugging or rather stay as a panic with a stack trace.

cmd/modelctl/common/endpoints.go
```diff
@@ -34,81 +56,146 @@ func ServerEndpoints(ctx *Context) (map[string]string, error) {
...
+			}
+		case "ws":
+			entrypoint, err = serverWsEntrypoint(ctx, serverSettings)
+			if err != nil {
+				return nil, fmt.Errorf("constructing WebSocket entrypoint: %v", err)
+			}
+		case "ws+unix":
+			entrypoint, err = serverWsOverUnixSocketEntrypoint(ctx, serverSettings)
+			if err != nil {
+				return nil, fmt.Errorf("constructing WebSocket Unix entrypoint: %v", err)
```
imatrisciano: If `https` is supported, arguably `wss` should be supported as well

pkg/runtimes/validate.go
```diff
@@ -128,8 +130,24 @@ func (server Server) validate(name string) error {
...
+		}
+
+		if parsed.Scheme != "" || parsed.Host != "" || parsed.Opaque != "" {
+			return fmt.Errorf("invalid base-path for server %s: must be a URL path, not a full URL", name)
+		}
+
+		if parsed.RawQuery != "" || parsed.Fragment != "" {
+			return fmt.Errorf("invalid base-path for server %s: query and fragment are not allowed", name)
+		}
 	}
```
imatrisciano, line 184: This validation function should verify that the value of `server.Protocol` is one of the supported ones, as defined in cmd/modelctl/common/endpoints.go

Since namespaces are used to identify a configuration group, should namespaces containing dots be rejected?

cmd/modelctl/commands/export-status.go
```diff
@@ -49,9 +49,17 @@ func (cmd *exportStatusCommand) run(_ *cobra.Command, args []string) error {
 		return fmt.Errorf("getting status: %v", err)
 	}
 
+	// Convert Entrypoints to Endpoints (extract URLs) for backward compatibility
```
imatrisciano: I would add a link to this PR for future reference

```suggestion
	// Convert Entrypoints to Endpoints (extract URLs) for backward compatibility,
	// see https://github.com/canonical/inference-snaps-cli/pull/412
```

imatrisciano (commented):
This looks solid, I've left a couple of comments

cmd/modelctl/commands/export-status.go
farshidtz: The comment was misleading. Nothing is being converted there for backward compatibility. The export object is filled with values taken from status.

imatrisciano (approved):
These are much needed changes to expand what's possible with inference snaps.

I've tested a local build and it's working as expected

